PR #6, opened 2019-04-17
Adding websockets command

jespino:


commands/init.go
```diff
@@ -41,3 +44,15 @@ func InitClient() (*model.Client4, error) {
...
 }
+
+func InitWebSocketClient() (*model.WebSocketClient, error) {
+	credentials, err := GetCurrentCredentials()
+	if err != nil {
+		return nil, err
+	}
+	client, appErr := model.NewWebSocketClient4(strings.Replace(credentials.InstanceUrl, "http", "ws", 1), credentials.AuthToken)
+	if appErr != nil {
+		return nil, errors.New("Unable to create the websockets connection")
```
mgdelacroix: Why not sending back the `appErr` or wrap it? I think this obscures the reason behind the error

jespino: Yes, let me check

jespino: Done!

PR #14, opened 2019-07-04
Adds a magefile to run the project's tasks

mgdelacroix:


mgdelacroix:
@cpanato good one! I'll do it in a follow up so it's isolated from the `magefile` itself, as the `Makefile` is still present and working :)

PR #21, opened 2019-10-04
With client refactor

mgdelacroix:
This PR renames the `command` variables into `cmd`, and refactors the `Client` initialisation to use the `withClient` decorator.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -9,14 +9,11 @@ import (
 )
 
 func withClient(fn func(c *model.Client4, cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) error {
-	c, err := InitClient()
-	if err != nil {
-		return func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		c, err := InitClient()
+		if err != nil {
 			return err
 		}
-	}
-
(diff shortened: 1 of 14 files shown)


commands/init.go
```diff
@@ -5,8 +5,22 @@ import (
...
+	c, err := InitClient()
+	if err != nil {
+		return func(cmd *cobra.Command, args []string) error {
+			return err
+		}
+	}
+
+	return func(cmd *cobra.Command, args []string) error {
+		return fn(c, cmd, args)
+	}
```
jespino: This is going to execute the initialization of the client for each method created, I think I prefer something like>

```suggestion
return func(cmd *cobra.Command, args []string) error {
	c, err := InitClient()
	if err != nil {
		return err
	}
	return fn(c, cmd, args)
}
```

mgdelacroix: Great catch!! Will change :)

mgdelacroix:
@streamer45 it's basically a way to generalise the initialisation of the client. Plenty of commands assume that you have a valid working connection, so those don't need a specific path to initialise and check that connection, hence the decorator.

Basically it avoids having

```go
c, err := InitClient()
if err != nil {
	return err
}
```

at the beginning of all those commands

PR #22, opened 2019-10-06
Replaces inline struct by a map[string]interface{}

mgdelacroix:


commands/channel.go
```diff
@@ -388,10 +388,10 @@ func searchChannelCmdF(c *model.Client4, cmd *cobra.Command, args []string) erro
 		var response *model.Response
 		channel, response = c.GetChannelByName(args[0], team.Id, "")
 		if response.Error != nil || channel == nil {
-			data := struct {
-				Channel string
-				Team    string
-			}{args[0], teamArg}
+			data := map[string]interface{}{
```
jespino: probably you can change it to `map[string]string` to be more concrete, which is the reasoning behind this change?

mgdelacroix: I was aiming for a broad generic solution, actually based on the `Fields` approach by logrus, that's why I went for `map[string]interface{}`, but you're right in the sense that we could go for a more specific solution in each case. I'm really 0/5 on picking one

PR #26, opened 2019-10-13
[MM-19065] Add auth renew command

mgdelacroix:
#### Summary

Adds an `mmctl auth renew SERVER` command that allows the user to quickly renew their credentials for a given server without the need to input again all the details of a server.


cpanato:
@lindalumitchell dont know if this appears in your notification and also dont know if you need to review that. I think will be good to at least be aware of the new feature :)

lindalumitchell:
Thanks @cpanato, I hadn't seen this one. 

Is this something that should/could have associated tests of some sort?

PR #31, opened 2019-10-27
Adds a testify suite for command testing

mgdelacroix:
#### Summary

This PR adds a couple of methods to the `printer` package that allow to capture the output during testing and reset the cummulated information of the `printer`. It wraps as well the `mmctl` unit tests with a `testify` suite, allowing to generalise the setup and teardown process of the tests, improving their readability and making them easier to write.


commands/user_test.go
```diff
@@ -0,0 +1,55 @@
...
+			Times(1)
+
+		err := searchUserCmdF(s.client, &cobra.Command{}, []string{emailArg})
+		s.Require().Nil(err)
+		s.Require().Equal(&mockUser, printer.GetLines()[0])
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+
+	s.Run("Search for a nonexistent user", func() {
+		printer.Clean()
```
mgdelacroix: I'm a little torn here. It seems that none of the methods of `suite` would allow us to hook at the beginning/end of a `s.Run`, so there is no way to seamlessly reset the `printer` state between subtests.

A possible solution is this, if you are using subtests, you have to manually reset the `printer` state. Another solution would be that the `printer.GetLines` and `printer.GetErrorLines` methods would reset the lines as well, maybe renamed as `printer.GetLinesAndClean`.

Any strong opinion on this?

mgdelacroix: After thinking about this for a while, we are going to stumble upon cases of subtests that don't check a type of lines, hence the next subtest will need to call `printer.Clean` anyways.

printer/printer.go
```diff
@@ -71,7 +74,24 @@ func Flush() {
 	}
 }
 
+// Resets the printer's cummulated lines
+func Clean() {
+	printer.Lines = []interface{}{}
+	printer.ErrorLines = []interface{}{}
+}
+
+// Returns the printer's cummulated lines
```
grundleborg:
```suggestion
// Returns the printer's accumulated lines
```

printer/printer.go
```diff
@@ -71,7 +74,24 @@ func Flush() {
 	}
 }
 
+// Resets the printer's cummulated lines
```
grundleborg:
```suggestion
// Resets the printer's accumulated lines
```

printer/printer.go
```diff
@@ -71,7 +74,24 @@ func Flush() {
...
+	printer.Lines = []interface{}{}
+	printer.ErrorLines = []interface{}{}
+}
+
+// Returns the printer's cummulated lines
+func GetLines() []interface{} {
+	return printer.Lines
+}
+
+// Returns the printer's cummulated error lines
```
grundleborg:
```suggestion
// Returns the printer's accumulated error lines
```

PR #34, opened 2019-11-01
[MM-19066] Added `user list` command

njkevlani:
Fixes https://github.com/mattermost/mattermost-server/issues/12965


commands/user.go
```diff
@@ -78,6 +78,14 @@ var SearchUserCmd = &cobra.Command{
 	RunE:    withClient(searchUserCmdF),
 }
 
+var ListUsersCmd = &cobra.Command{
+	Use:     "list",
+	Short:   "List users",
+	Long:    "List all the users",
+	Example: "  user list",
+	RunE:    withClient(listUsersCmdF),
```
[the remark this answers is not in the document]
njkevlani: There are three args possible for this command: `--page`, `--per-page` and `--all`. (4th `--format`)

Let me know if I'm missing something here :)

njkevlani:
> Would you be open to add some unit tests for these PR? You can take inspiration on the [search channel command test](https://github.com/mattermost/mmctl/blob/master/commands/channel_test.go)

Sure, will do that.

commands/user.go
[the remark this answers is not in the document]
mgdelacroix: Those are cobra flags, and they don't count for the `Args` check :)

njkevlani:
@mgdelacroix 

Tests added. Let me know if this seems fine :)

commands/user_test.go
```diff
@@ -52,3 +52,69 @@ func (s *MmctlUnitTestSuite) TestSearchUserCmd() {
...
+		s.Require().Equal(&mockUser, printer.GetLines()[0])
+	})
+
+	s.Run("Listing all the users", func() {
+		emailArg := "example2@example.com"
+		mockUser := model.User{Username: "ExampleUser2", Email: emailArg}
+		perPage := 200
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Bool("--all", true, "")
```
mgdelacroix: There is a typo in the flag name, that is causing `showAll` to be set to `false`:

```suggestion
		cmd.Flags().Bool("all", true, "")
```

Setting the flag to true causes another call to the `GetUsers` method that needs to be added to the expectations

commands/user_test.go
```diff
@@ -52,3 +52,76 @@ func (s *MmctlUnitTestSuite) TestSearchUserCmd() {
...
+		emailArg := "example@example.com"
+		mockUser := model.User{Username: "ExampleUser", Email: emailArg}
+		page := 0
+		perPage := 1
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Int("per-page", perPage, "")
+		cmd.Flags().Int("page", page, "")
+
+		printer.Clean()
```
mgdelacroix: just one last nit: for clarity, would you be open to move the `printer.Clean()` calls to be the first thing done at the beginning of each test case, just right after the `s.Run`?

I think they can be missed if they are in the middle of the test rather than at first.

commands/user_test.go
```diff
@@ -52,3 +52,76 @@ func (s *MmctlUnitTestSuite) TestSearchUserCmd() {
...
+	s.Run("Listing all the users", func() {
+		emailArg := "example2@example.com"
+		mockUser := model.User{Username: "ExampleUser2", Email: emailArg}
+		perPage := 200
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Bool("all", true, "")
+		cmd.Flags().Int("per-page", perPage, "")
+
+		printer.Clean()
```
mgdelacroix: Same here

commands/user_test.go
```diff
@@ -52,3 +52,76 @@ func (s *MmctlUnitTestSuite) TestSearchUserCmd() {
...
+
+	s.Run("Try to listing all the users when there are no uses in store", func() {
+		page := 0
+		perPage := 1
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Int("per-page", perPage, "")
+		cmd.Flags().Int("page", page, "")
+
+		printer.Clean()
```
mgdelacroix: Same here

hanzei:
@njkevlani Would you please merge `master` into the branch, when you have the time?

njkevlani:
@hanzei master branch is merged into this branch.

njkevlani:
ping @mkraft

commands/user.go
```diff
@@ -315,3 +329,23 @@ auth_service: {{.AuthService}}`
...
 	return nil
 }
+
+func listUsersCmdF(c client.Client, command *cobra.Command, args []string) error {
+	page, _ := command.Flags().GetInt("page")
+	perPage, _ := command.Flags().GetInt("per-page")
+	showAll, _ := command.Flags().GetBool("all")
+
+	tpl := `{{.Id}}: {{.Username}} ({{.Email}})`
+	for users, _ := c.GetUsers(page, perPage, ""); users != nil && len(users) > 0; users, _ = c.GetUsers(page, perPage, "") {
```
mkraft: What's the reason we're using the blank identifier for these `GetUsers` errors?

hanzei:
@njkevlani I just want to check if you have any questions about the provided feedback. The PR is almost over the finish line :tada:

njkevlani:
@hanzei, there are couple of reviews pending [here](https://github.com/mattermost/mmctl/pull/34#discussion_r356397343) and [here](https://github.com/mattermost/mmctl/pull/34#discussion_r344471083).

If you guys are okay with them, then it looks good to me.

hanzei:
I think the only outstanding review is a question from `@mkraft': https://github.com/mattermost/mmctl/pull/34#discussion_r355983168

`@mgdelacroix` already approved the PR.

njkevlani:
Then it looks good to me :)

njkevlani:
Any status update? @hanzei @lindy65 @mkraft @mgdelacroix

lindy65:
Thanks @njkevlani - I've passed this for QA as it will be tested post-merge. 👍

commands/user.go
jasonblais: @njkevlani would you be open to help with @mkraft's question above? Thanks!

hanzei:
@agnivade Would you please review this PR given that @mkraft is OOO

commands/user_test.go
```diff
@@ -870,3 +870,79 @@ func (s *MmctlUnitTestSuite) TestResetUserMfaCmd() {
...
+		err := listUsersCmdF(s.client, cmd, []string{})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(&mockUser, printer.GetLines()[0])
+	})
+
+	s.Run("Listing all the users", func() {
+		printer.Clean()
+
+		emailArg := "example2@example.com"
```
agnivade: This is a separate test. Let's use the same email and username for consistency.

commands/user.go
```diff
@@ -91,6 +100,10 @@ func init() {
 	UserCreateCmd.Flags().String("locale", "", "Optional. The locale (ex: en, fr) for the new user account.")
 	UserCreateCmd.Flags().Bool("system_admin", false, "Optional. If supplied, the new user will be a system administrator. Defaults to false.")
 
+	ListUsersCmd.Flags().Int("page", 0, "Start page for list of users")
+	ListUsersCmd.Flags().Int("per-page", 200, "Number of users to be fetched")
+	ListUsersCmd.Flags().Bool("all", false, "Fetch all users. Will ignore --page and --per-page")
```
agnivade: I don't see how this ignores `--page` and `--per-page`. From what I understand, if `all` is set, it just increments the page number and keeps on getting new pages. That's the only difference. Can you clarify ?

Also, there is no need for it to ignore `per-page`. It should work fine with all. And that's what it seems to do. Just that it should ignore page, which it does not seem to do.

commands/user_test.go
```diff
@@ -870,3 +870,79 @@ func (s *MmctlUnitTestSuite) TestResetUserMfaCmd() {
...
+			Return([]*model.User{}, &model.Response{Error: nil}).
+			Times(1)
+
+		err := listUsersCmdF(s.client, cmd, []string{})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(&mockUser, printer.GetLines()[0])
+	})
+
+	s.Run("Try to listing all the users when there are no uses in store", func() {
```
agnivade: ... list all users .. no users in ..

commands/user_test.go
```diff
@@ -870,3 +870,79 @@ func (s *MmctlUnitTestSuite) TestResetUserMfaCmd() {
...
+		s.client.
+			EXPECT().
+			GetUsers(0, perPage, "").
+			Return([]*model.User{&mockUser}, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			GetUsers(1, perPage, "").
+			Return([]*model.User{}, &model.Response{Error: nil}).
```
agnivade: Please return another user here. I would like see 2 separate users from 2 pages.

commands/user_test.go
```diff
@@ -870,3 +870,79 @@ func (s *MmctlUnitTestSuite) TestResetUserMfaCmd() {
...
+		emailArg := "example@example.com"
+		mockUser := model.User{Username: "ExampleUser", Email: emailArg}
+		page := 0
+		perPage := 1
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Int("per-page", perPage, "")
+		cmd.Flags().Int("page", page, "")
+
+		s.client.
```
agnivade: Please return 2 pages of users, each page having 1 user. And verify that both users are returned.

commands/user.go
```diff
@@ -315,3 +329,23 @@ auth_service: {{.AuthService}}`
 
 	return nil
 }
+
+func listUsersCmdF(c client.Client, command *cobra.Command, args []string) error {
+	page, _ := command.Flags().GetInt("page")
```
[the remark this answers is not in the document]
agnivade: No need to wrap them up with an error message. The error will itself contain all we need
- `fmt.Errorf("flag accessed but not defined: %s", name)`
- `fmt.Errorf("trying to get %s value of flag of type %s", ftype, flag.Value.Type())`

commands/user.go
agnivade: Sounds good. Just a small nit: it should be `--page flag will be ignored if provided`. Thanks !

agnivade:
> Won't it be similar to the first test case (Listing users with paging)?

I don't believe so. That test case only has 1 page, and the client also requests for that page. How do we know that the server isn't returning all pages ?

What I am suggesting is to have 3 pages, and have the client return the last page. That verifies that the code can indeed return the correct page. Does that make sense ?

commands/user_test.go
```diff
@@ -870,3 +872,120 @@ func (s *MmctlUnitTestSuite) TestResetUserMfaCmd() {
...
+			EXPECT().
+			GetUsers(page, perPage, "").
+			Return(nil, &model.Response{Error: &mockError}).
+			Times(1)
+
+		err := listUsersCmdF(s.client, cmd, []string{})
+		s.Require().NotNil(err)
+		s.Require().Equal(mockErrorW.Error(), err.Error())
```
agnivade: You can just do `s.EqualError(err, mockErrorW.Error())`

client/client.go
```diff
@@ -60,4 +60,5 @@ type Client interface {
 	DeleteCommand(commandId string) (bool, *model.Response)
 	GetConfig() (*model.Config, *model.Response)
 	SyncLdap() (bool, *model.Response)
+	GetUsers(page int, perPage int, etag string) ([]*model.User, *model.Response)
```
agnivade: Oh I missed this. Please combine page and perPage into 1 - `page, perPage int,`

commands/user.go
```diff
@@ -315,3 +329,45 @@ auth_service: {{.AuthService}}`
...
 	return nil
 }
+
+func listUsersCmdF(c client.Client, command *cobra.Command, args []string) error {
+	// TODO What to do with err?
+	page, errp := command.Flags().GetInt("page")
+	if errp != nil {
+		return errors.Wrap(errp, "Could not parse --page")
+	}
+	perPage, errpp := command.Flags().GetInt("per-page")
```
agnivade: I think we can reuse the same error variable. There's no need to name a new one.

commands/user.go
```diff
@@ -315,3 +329,45 @@ auth_service: {{.AuthService}}`
...
+		return errors.Wrap(errs, "Could not parse --all")
+	}
+
+	if showAll {
+		page = 0
+	}
+
+	tpl := `{{.Id}}: {{.Username}} ({{.Email}})`
+
+	users, response := c.GetUsers(page, perPage, "")
```
agnivade: I still don't understand why we are calling this twice ? Can't we have a single loop where we are calling this once ?

commands/user.go
```diff
@@ -315,3 +329,45 @@ auth_service: {{.AuthService}}`
...
+	}
+	for ; users != nil && len(users) > 0; users, response = c.GetUsers(page, perPage, "") {
+		if response.Error != nil {
+			return errors.Wrap(response.Error, "Failed to fetch users")
+		}
+		for _, user := range users {
+			printer.PrintT(tpl, user)
+		}
+
+		if !showAll {
```
agnivade: I don't think this is needed anymore now the `page = 0` gets set when showAll is true. This will not print any more pages in the normal mode (when page and perPage is passed)

commands/user.go
njkevlani: The condition in for loop is:
```users != nil && len(users) > 0```
If server returns error on first call, `users` will be `nil`. So the part inside loop will never execute and `nil` will be returned.

One option is to check for response.Error before [return statement](https://github.com/mattermost/mmctl/pull/34/files#diff-dbf54b6a188b305f65470b9f52908c27R372).

commands/user.go
njkevlani: If we ignore this condition, then all users will be printed starting from page `page` even if `showAll` is `false`.

agnivade: I thought that was the intention ? Otherwise we should not mention "Start page", because that means the start offset from where all pages would be printed.

@mgdelacroix - Can you clarify ?

commands/user.go
agnivade: Apologies for the confusion. By "calling this twice", I meant why are we writing the code invoking `GetUsers` twice. I get that the method actually gets invoked once, but why do we need to write it twice ? Can't we just `c.GetUsers` once in the code ?

Is that clear ?

commands/user.go
mgdelacroix: The intent was for the command to either show all the users when the `--all` flag is true, or one page of the user list. I agree with ditching "start page" in favor of something clearer

agnivade: Maybe something like "Page number to fetch for the list of users" ?

agnivade:
We are nearly there. Just one outstanding [comment](https://github.com/mattermost/mmctl/pull/34/files#r365790935) to clarify. And please fix the merge conflicts. Thanks !

njkevlani:
> We are nearly there. Just one outstanding [comment](https://github.com/mattermost/mmctl/pull/34/files#r365790935) to clarify.

Does [this reply](https://github.com/mattermost/mmctl/pull/34/files#r365797346) answer the question in linked comment?

agnivade:
That doesn't link to a comment for me. :( Feel free to just paste the comment.

njkevlani:
> Sorry, perhaps there is some miscommunication amongst us. I meant why do we need to call the GetUsers method twice. Does something like this work ?
> 
> ```go
> for {
> 	users, res := c.GetUsers(page, perPage, "")
> 	if res.Error != nil {
> 		return errors.Wrap(...)
> 	}
> 	if len(users) == 0 {
> 		break
> 	}
> 	for _, user := range users {
> 		printer.PrintT(tpl, user)
> 	}
> 
> 	if !showAll {
> 		break
> 	}
> 	page++	
> }
> ```

Yes, this would work fine. I guess the difference in this approach any my previous code is that,
  - In this approach: `for { ... }` loop is used. Initialization / assignment occurs inside loop body
  - In my previous approach: `for init; condition; post { ... }` initialization / assignment were part of loop,

Anyway, I have made changes in code. Let me know if all look good.

njkevlani:
Thanks for feedback on this, got to learn some things :) @agnivade

njkevlani:
@mgdelacroix any updates?

mgdelacroix:
Thanks a lot for all the changes you've implemented! Can you please run `make docs` and update the PR to fix the CI?

njkevlani:
> @njkevlani thanks a lot for all the changes you've implemented! Can you please run `make docs` and update the PR to fix the CI?

Done :)

PR #40, opened 2019-11-12
[MM-20031] Add unit tests to "userDeactivateCmd" to use testify

phillipahereza:
This PR adds unit tests to the `userDeactivateCmd` command

Ticket: [Jira Ticket](https://mattermost.atlassian.net/browse/MM-20031)

Fixes https://github.com/mattermost/mattermost-server/issues/13047


commands/user_test.go
```diff
@@ -52,3 +52,284 @@ func (s *MmctlUnitTestSuite) TestSearchUserCmd() {
...
+			Return(nil, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			GetUserByUsername(argsDelete[2], "").
+			Return(&argUsers[2], &model.Response{Error: nil}).
+			Times(1)
+
+		for i := 0; i < len(argsDelete); i++ {
```
mgdelacroix: nit: you can iterate over `argsUsers` to improve readability:

```suggestion
		for _, user := range argUsers {
```

commands/user_test.go
```diff
@@ -52,3 +52,284 @@ func (s *MmctlUnitTestSuite) TestSearchUserCmd() {
...
+
+		s.client.
+			EXPECT().
+			GetUserByEmail(emailArg, "").
+			Return(&mockUser, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			DeleteUser(mockUser.Id).
```
marianunez: Just curious, why do we call `DeleteUser` before calling the `userDeactivateCommand`?

jasonblais:
Hey @phillipahereza, can you help with Maria's question above? Seems that's the only outstanding question on this PR. Thanks for your contribution!

commands/user_test.go
phillipahereza: Sorry for the late response but I assumed the `userDeactivateCommand` is a wrapper around `DeleteUser` so we don't call it before calling `userDeactivateCommand` but rather during the `userDeactivateCommand`.

PR #47, opened 2019-11-16
[MM-20189] Add unit tests to "updateUserEmailCmd" to use testify

enolal826:
**Summary**
Unit testing added for UpdateUserEmailCmd

Fixes [mattermost/mattermost-server#13109](https://github.com/mattermost/mattermost-server/issues/13109)


cpoile:
Thanks @enolal826 for the contribution!
@mgdelacroix the title of the HW ticket is a bit confusing for me -- are we supposed to be moving to testify require and asserts?

mgdelacroix:
> Are we supposed to be moving to testify require and asserts?

The goal of the ticket is to create the tests for the `updateUserEmailCmd`. Those tests must use `testify` as per the examples on the blogpost in the description, but I agree on the title being a bit confusing with the recent `testify` campaign (it has been changed already for future tickets to remove the `testify` reference).

commands/user_test.go
```diff
@@ -190,3 +190,189 @@ func (s *MmctlUnitTestSuite) TestUserCreateCmd() {
...
 	})
 }
+
+func (s *MmctlUnitTestSuite) TestUpdateUserEmailCmd() {
+	s.Run("Two arguments are not provided", func() {
+		command := cobra.Command{}
+
+		error := updateUserEmailCmdF(s.client, &command, []string{})
+
+		s.Require().Equal("Expected two arguments. See help text for details.", error.Error())
```
mgdelacroix: I would add as well `s.Len` checks for the amount of lines on `printer.GetLines()` and `printer.GetErrorLines()` slices, to ensure that the printed output is the expected.

When checking the message of an error, you can use [`require.EqualError`](https://godoc.org/github.com/stretchr/testify/require#EqualError):

```suggestion
		s.Require().EqualError(error, "Expected two arguments. See help text for details.")
```

commands/user_test.go
```diff
@@ -190,3 +190,189 @@ func (s *MmctlUnitTestSuite) TestUserCreateCmd() {
...
+		s.client.
+			EXPECT().
+			GetUserByUsername(userArg, "").
+			Return(&mockUser, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			UpdateUser(&mockUser).
+			Return(&mockUser, &model.Response{Error: nil}).
```
mgdelacroix: This mock should return an `updatedUser` instance of the user with the new email, so we can check that the response is correct instead of returning the original user.

commands/user_test.go
```diff
@@ -190,3 +190,189 @@ func (s *MmctlUnitTestSuite) TestUserCreateCmd() {
...
+			Times(1)
+
+		error := updateUserEmailCmdF(s.client, &command, []string{userArg, emailArg})
+
+		s.Require().Nil(error)
+		s.Require().Equal(&mockUser, printer.GetLines()[0])
```
mgdelacroix: Here we should check for the `updatedUser` instance instead

commands/user_test.go
```diff
@@ -190,3 +190,189 @@ func (s *MmctlUnitTestSuite) TestUserCreateCmd() {
...
```
mgdelacroix: Same `updatedUser` check here

commands/user_test.go
```diff
@@ -190,3 +190,189 @@ func (s *MmctlUnitTestSuite) TestUserCreateCmd() {
...
```
mgdelacroix: Same `updatedUser` check here

commands/user_test.go
```diff
@@ -190,3 +190,189 @@ func (s *MmctlUnitTestSuite) TestUserCreateCmd() {
 		s.Require().Equal("Unable to update user roles. Error: : Remote error, ", error.Error())
 	})
 }
+
+func (s *MmctlUnitTestSuite) TestUpdateUserEmailCmd() {
+	s.Run("Two arguments are not provided", func() {
+		command := cobra.Command{}
```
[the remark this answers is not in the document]
enolal826: Done! Thanks

commands/user_test.go
enolal826: Totally right, updated. Thanks!

commands/user_test.go
enolal826: I have changed the EqualError method, but I'm not sure if it's worth it adding the printer checks. There is no printing on error cases, so we should be doing something like:

`s.Require().Len(printer.GetLines(), 0)`

commands/user_test.go
enolal826: Updated!

commands/user_test.go
enolal826: Updated!

commands/user_test.go
enolal826: Updated!

cpoile:
@enolal826 It looks like there is a merge conflict -- can you manually resolve that?

enolal826:
@cpoile  Updated with master, thanks!

PR #53, opened 2019-11-18
Bump go version to 1.13 and fix dependency versions in go.mod

mgdelacroix:
#### Summary

After removing workaround for old go versions, explicitly state the need of using `go1.13`, fixing direct dependency versions, cleaning `go.mod` and modifying build tools to avoid unnecessary updating `go.mod` when downloading `shadow`.


go.mod
```diff
@@ -1,15 +1,13 @@
...
 
-go 1.12
+go 1.13
 
 require (
-	github.com/cpuguy83/go-md2man v1.0.10 // indirect
-	github.com/golang/mock v1.2.0
-	github.com/magefile/mage v1.8.0
+	github.com/golang/mock v1.3.1
+	github.com/magefile/mage v1.9.0
```
hanzei: Why is `mage` part of `go.mod`? We are not importing it anywhere.

mgdelacroix: It is being imported inside `magefile.go`, so is being taking into account by the vendor command

hanzei: I see :+1:

PR #55, opened 2019-11-19
[MM-19912] Add end to end tests to mmctl

mgdelacroix:
#### Summary

This PR adds an example of an end to end test and the necessary infrastructure for the CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ executors:
   default:
     working_directory: ~/go/src/github.com/mattermost/mmctl
     docker:
-      - image: circleci/golang:1.12
+      - image: circleci/golang:1.13
 
 jobs:
   lint:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 executors:
   default:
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -13,12 +13,92 @@ jobs:
...
+          command: |
+            ulimit -n 8096
+            cd ~/go/src/github.com/mattermost/mattermost-server
+
+            docker run --net circleci_mm-test \
+              --name mattermost_circleci \
+              --env MM_CONFIG='mysql://mmuser:mostest@tcp(mysql:3306)/mattermost_test?charset=utf8mb4,utf8' \
+              --env MM_TEAMSETTINGS_MAXUSERSPERTEAM=500 \
+              -d \
+              mattermost/mattermost-enterprise-edition:5.14.0-rc4
```
cpanato:
```suggestion
              mattermost/mattermost-enterprise-edition:master
```
will get the latest master available

cpanato: or do you want to test only on 5.14 release?

mgdelacroix: Nope, that's definitely a good catch!

PR #63, opened 2019-11-29
Add two new commands for config: set and reset

ethervoid:
This PR adds two new options (set and reset) to the config option for mmctl

This PR depends on #64 so we need to merge that PR before this one

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -237,11 +237,12 @@ func configResetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	if _, res := c.UpdateConfig(config); res.Error != nil {
+	newConfig, res := c.UpdateConfig(config)
+	if res.Error != nil {
 		return res.Error
 	}
 
-	printer.Print("Value/s reset successfully")
+	printer.PrintT("Value/s reset successfully", newConfig)
 	return nil
 }
 
--- a/commands/config.go
+++ b/commands/config.go
@@ -81,7 +81,12 @@ func getValue(path []string, obj interface{}) (interface{}, bool) {
 }
 
 func setValue(path []string, obj reflect.Value, newValue interface{}) error {
-	val := obj.FieldByName(path[0])
+	var val reflect.Value
+	if obj.Kind() == reflect.Struct {
+		val = obj.FieldByName(path[0])
+	} else {
+		val = obj
+	}
 
 	if val.Kind() == reflect.Invalid {
 		return errors.New("Selected path object is not valid")
@@ -90,44 +95,43 @@ func setValue(path []string, obj reflect.Value, newValue interface{}) error {
 	if len(path) > 1 && val.Kind() == reflect.Struct {
 		return setValue(path[1:], val, newValue)
 	} else if len(path) == 1 {
-		if val.Kind() != reflect.Ptr && val.Kind() != reflect.Slice && val.Kind() != reflect.Struct {
-			return errors.New("Value is not modifiable")
-		} else if val.Kind() == reflect.Ptr {
-			switch val.Elem().Kind() {
+		if val.Kind() == reflect.Ptr {
+			return setValue(path, val.Elem(), newValue)
+		} else if val.Kind() == reflect.Struct {
+			val.Set(reflect.ValueOf(newValue))
+			return nil
+		} else if val.Kind() == reflect.Slice {
+			if val.Type().Elem().Kind() != reflect.String {
+				return errors.New("Unsupported type of slice")
+			}
+			val.Set(reflect.ValueOf(newValue))
+			return nil
+		} else {
+			switch val.Kind() {
 			case reflect.Int:
 				v, err := strconv.ParseInt(newValue.(string), 10, 64)
 				if err != nil {
 					return errors.New("Target value is of type Int and provided value is not")
 				}
-				val.Elem().SetInt(v)
+				val.SetInt(v)
 				return nil
 			case reflect.String:
-				val.Elem().SetString(newValue.(string))
+				val.SetString(newValue.(string))
 				return nil
 			case reflect.Bool:
 				v, err := strconv.ParseBool(newValue.(string))
 				if err != nil {
 					return errors.New("Target value is of type Bool and provided value is not")
 				}
-				val.Elem().SetBool(v)
+				val.SetBool(v)
 				return nil
 			default:
 				return errors.New("Target value type is not supported")
 			}
-		} else if val.Kind() == reflect.Struct {
-			val.Set(reflect.ValueOf(newValue))
-			return nil
-		} else if val.Type().Elem().Kind() == reflect.String {
-			val.Set(reflect.ValueOf(newValue))
-			return nil
-		} else {
-			return errors.New("Unsupported type of slice")
 		}
 	} else {
 		return errors.New("Path object type is not supported")
 	}
-
-	return nil
 }
 
 func setConfigValue(path []string, config *model.Config, newValue []string) error {
--- a/commands/config.go
+++ b/commands/config.go
@@ -191,11 +191,12 @@ func configSetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
 	if err := setConfigValue(path, config, args[1:]); err != nil {
 		return err
 	}
-	if _, res := c.UpdateConfig(config); res.Error != nil {
+	newConfig, res := c.UpdateConfig(config)
+	if res.Error != nil {
 		return res.Error
 	}
 
-	printer.Print("Value changed successfully")
+	printer.PrintT("Value changed successfully", newConfig)
 	return nil
 }
 
--- a/commands/config.go
+++ b/commands/config.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"encoding/json"
 	"errors"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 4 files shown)


commands/config.go
```diff
@@ -78,6 +178,69 @@ func configGetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
...
+		return res.Error
+	}
+
+	printer.Print("Value/s reset successfully")
+	return nil
+}
+
 func configShowCmdF(c client.Client, cmd *cobra.Command, args []string) error {
 	printer.SetSingle(true)
 	printer.SetFormat(printer.FORMAT_JSON)
```
ethervoid: Is this right? Our help says default value for `--format` is plain but we're showing only JSON in this command

mgdelacroix: In this specific case, the way to show the documentation is the same in both formats (a JSON entry) hence the fixed use of JSON.

We could potentially have another visualization for the "plain" mode if there is one that makes sense

ethervoid:
@mgdelacroix gentle remainder for the PR :)

mgdelacroix:
Apologies for the waiting @ethervoid, I'll review it through the day :)

commands/config.go
```diff
@@ -78,6 +178,69 @@ func configGetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
...
+
+func configResetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
+	confirmFlag, _ := cmd.Flags().GetBool("confirm")
+
+	if !confirmFlag && len(args) > 0 {
+		var confirmResetAll string
+		confirmationMsg := fmt.Sprintf(
+			"Are you sure you want to reset %s to their default value? (YES/NO): ",
+			args[0])
+		printer.Print(confirmationMsg)
```
mgdelacroix: Confirmation message should be printed directly with `fmt`, as we don't want it to be included as part of the JSON output and we do want it to be interactive (and `--format json` would accumulate it instead of printing it)

commands/config.go
```diff
@@ -78,6 +178,69 @@ func configGetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
...
+		fmt.Scanln(&confirmResetAll)
+		if confirmResetAll != "YES" {
+			printer.Print("Reset operation aborted")
```
mgdelacroix: Same here, tho it's just about to exit, for consistency

commands/config.go
```diff
@@ -78,6 +178,69 @@ func configGetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
...
+	for _, arg := range args {
+		path, err := parseConfigPath(arg)
+		if err != nil {
+			return err
+		}
+		defaultValue, ok := getValue(path, *defaultConfig)
+		if !ok {
+			return errors.New("Invalid key")
+		}
+		resetConfigValue(path, config, defaultValue)
```
mgdelacroix: Here the error should be handled before trying to update the config, as we are risking a partial update

commands/config_test.go
```diff
@@ -0,0 +1,392 @@
...
+		args := []string{"SqlSettings.WrongKey", "test1"}
+		inputConfig := &model.Config{}
+		inputConfig.SetDefaults()
+
+		s.client.
+			EXPECT().
+			GetConfig().
+			Return(defaultConfig, &model.Response{Error: nil}).
+			Times(1)
+		s.client.
```
mgdelacroix: Mocking this interaction is not needed, as the `gomock` library will fail on any interaction with the client that is not mocked already

commands/config_test.go
```diff
@@ -0,0 +1,392 @@
...
+		args := []string{"WrongKey"}
+		defaultConfig := &model.Config{}
+		defaultConfig.SetDefaults()
+
+		s.client.
+			EXPECT().
+			GetConfig().
+			Return(defaultConfig, &model.Response{Error: nil}).
+			Times(1)
+		s.client.
```
mgdelacroix: Same here

commands/config.go
```diff
@@ -78,6 +178,69 @@ func configGetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	if err := setConfigValue(path, config, args[1:]); err != nil {
+		return err
+	}
+	if _, res := c.UpdateConfig(config); res.Error != nil {
+		return res.Error
+	}
+
+	printer.Print("Value changed successfully")
```
mgdelacroix: In case of success, I'd suggest to print the whole successful config if the `--format json` is enabled, capturing the return value as `newConfig` (for example) and then printing something like:

```suggestion
	printer.PrintT("Value changed successfully", newConfig)
```

So plain output would just be a confirmation message, and JSON output would yield the newly saved config

commands/config.go
```diff
@@ -78,6 +178,69 @@ func configGetCmdF(c client.Client, cmd *cobra.Command, args []string) error {
...
+		if !ok {
+			return errors.New("Invalid key")
+		}
+		resetConfigValue(path, config, defaultValue)
+	}
+	if _, res := c.UpdateConfig(config); res.Error != nil {
+		return res.Error
+	}
+
+	printer.Print("Value/s reset successfully")
```
mgdelacroix: Same here, we should capture the `newConfig` return value from the call and replace:

```suggestion
	printer.PrintT("Value/s reset successfully", newConfig)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
+			EXPECT().
+			GetConfig().
+			Return(outputConfig, &model.Response{Error: nil}).
+			Times(1)
+
+		err := configGetCmdF(s.client, &cobra.Command{}, args)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0].([]string), []string{})
```
mgdelacroix: We should be able to remove the casting here:

```suggestion
		s.Require().Equal(printer.GetLines()[0], []string{})
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
+		s.client.
+			EXPECT().
+			GetConfig().
+			Return(outputConfig, &model.Response{Error: nil}).
+			Times(1)
+
+		err := configGetCmdF(s.client, &cobra.Command{}, args)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0].(model.SqlSettings), sqlSettings)
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], sqlSettings)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
+		s.client.
+			EXPECT().
+			UpdateConfig(inputConfig).
+			Return(inputConfig, &model.Response{Error: nil}).
+			Times(1)
+
+		err := configSetCmdF(s.client, &cobra.Command{}, args)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0].(*model.Config), inputConfig)
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], inputConfig)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], inputConfig)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], inputConfig)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], inputConfig)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
+			UpdateConfig(defaultConfig).
+			Return(defaultConfig, &model.Response{Error: nil}).
+			Times(1)
+
+		resetCmd := &cobra.Command{}
+		resetCmd.Flags().Bool("confirm", true, "")
+		err := configResetCmdF(s.client, resetCmd, args)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0].(*model.Config), defaultConfig)
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], defaultConfig)
```

commands/config_test.go
```diff
@@ -0,0 +1,382 @@
...
+			Return(defaultConfig, &model.Response{Error: nil}).
+			Times(1)
+
+		resetCmd := &cobra.Command{}
+		resetCmd.Flags().Bool("confirm", true, "")
+		resetCmd.ParseFlags([]string{"confirm"})
+		err := configResetCmdF(s.client, resetCmd, args)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0].(*model.Config), defaultConfig)
```
mgdelacroix: Same:

```suggestion
		s.Require().Equal(printer.GetLines()[0], defaultConfig)
```

PR #67, opened 2019-12-03
[GH-13278] Port the "user activate" subcommand from Mattermost CLI to mmctl

larkox:
Summary:
Port the "user activate" subcommand from Mattermost CLI to mmctl

Link:
Fixes https://github.com/mattermost/mattermost-server/issues/13278


commands/user.go
```diff
@@ -104,30 +114,47 @@ func init() {
 	RootCmd.AddCommand(UserCmd)
 }
 
-func deactivateUsers(c client.Client, userArgs []string) {
+func userActivateCmdF(c client.Client, command *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return errors.New("Expected at least one argument. See help text for details.")
+	}
```
mgdelacroix: This check can be replaced using `Args: cobra.MinimumNArgs(1),` in the command struct

commands/user.go
```diff
@@ -104,30 +114,47 @@ func init() {
...
+}
 
-		if _, response := c.DeleteUser(user.Id); response.Error != nil {
-			printer.PrintError("Unable to deactivate user " + userArgs[i] + ". Error: " + response.Error.Error())
-		}
+func changeUserActiveStatus(c client.Client, user *model.User, userArg string, activate bool) error {
+	if user == nil {
+		return fmt.Errorf("Can't find user '%v'", userArg)
+	}
```
mgdelacroix: I'd prefer this check to be made right after getting the user, and for the `changeUserActiveStatus` function to trust the arguments so it's simpler and the errors are handled right where they occur. [3/5](https://docs.mattermost.com/process/training.html#out-of-5)

commands/user_test.go
```diff
@@ -9,6 +10,351 @@ import (
...
+			Times(1)
+
+		err := userActivateCmdF(s.client, &cobra.Command{}, []string{emailArg})
+		s.Require().NoError(err)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 1)
+		s.Require().Equal(fmt.Errorf("Unable to change activation status of user: %v", emailArg).Error(), printer.GetErrorLines()[0])
+	})
+
+	s.Run("Activate several users with unexistent ones and failure ones", func() {
```
mgdelacroix:
```suggestion
	s.Run("Activate several users with unexistent ones and failed ones", func() {
```

commands/user.go
```diff
@@ -104,30 +114,47 @@ func init() {
...
 	}
+	if _, response := c.UpdateUserActive(user.Id, activate); response.Error != nil {
+		return fmt.Errorf("Unable to change activation status of user: %v", userArg)
+	}
+
+	return nil
 }
 
 func userDeactivateCmdF(c client.Client, cmd *cobra.Command, args []string) error {
 	if len(args) < 1 {
```
mgdelacroix: It doesn't belong specifically to this ticket, but it would be great if you could apply the `Args: cobra.MinimumNArgs(1)` to this command as well

commands/user.go
```diff
@@ -104,30 +114,47 @@ func init() {
...
+		if err != nil {
+			printer.PrintError(err.Error())
 		}
+	}
+}
 
-		if _, response := c.DeleteUser(user.Id); response.Error != nil {
-			printer.PrintError("Unable to deactivate user " + userArgs[i] + ". Error: " + response.Error.Error())
-		}
+func changeUserActiveStatus(c client.Client, user *model.User, userArg string, activate bool) error {
```
mgdelacroix: I don't think we are using `c.DeleteUser` anywhere else. If so, would you mind removing it from the `Client` and updating the mocks?

commands/user.go
larkox: If this is the only workflow that is expected (changeUsersActiveStatus->changeUserActiveStatus) this work. If at any moment we decide to expose or use in other place the "changeUserActiveStatus" function, it may result in a panic. That is why I generally prefer to check things just before using them.

Nevertheless, if we want to enforce that changeUserActiveStatus to trust the arguments and expect a panic if someone or something is using it wrong, we can make the change.

mgdelacroix: For me is the other way around, if we retrieve something but don't check that we got a value, that could cause a panic. Now we are doing `changeUsersActiveStatus -> changeUserActiveStatus`, but if we decide to do something else with those users before (or even after) going through the function that checks the `nil`s, we would be operating with usafe values.

Checking something as part of getting it is safe, as it doesn't matter where you use it, you know that it has already been checked, hence the rest of the functions can trust that value (or check it for double safety if the functions are exposed as part of an API).

PR #71, opened 2019-12-07
[MM-20837] Add unit tests to the "ConfigGetCmd" mmctl command

artemigkh:
#### Summary
Added unit tests for the ConfigGetCmd mmctl command that covers the following test cases:
- Get a single nested config setting
- Get a block of config settings
- Assert Invalid Key error is returned when accessing nonexistent config field
- Assert error on client request is gracefully handled

#### Ticket Links
Jira: https://mattermost.atlassian.net/browse/MM-20837
Fixes mattermost/mattermost-server#13287

P.S - this repo didn't quite work for me out of the box, failing with the same errors CI is; I had to replace the import `"github.com/mattermost/mattermost-server/model"` in `license_test.go` with `"github.com/mattermost/mattermost-server/v5/model"` for it to build and have the tests run properly

### DIFF
--- a/commands/config_test.go
+++ b/commands/config_test.go
@@ -1,0 +1,105 @@
+package commands
+
+import (
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mmctl/printer"
+	"github.com/spf13/cobra"
+)
+
+func (s *MmctlUnitTestSuite) TestConfigGetCmdF() {
+	s.Run("Get nested config setting", func() {
+		printer.Clean()
+
+		configNameArg := "ServiceSettings.WebsocketSecurePort"
+		configValue := 443
+		mockConfig := &model.Config{
+			ServiceSettings: model.ServiceSettings{
+				WebsocketSecurePort: &configValue,
+			},
+		}
+
+		s.client.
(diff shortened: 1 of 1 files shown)


hanzei:
Hi @artemigkh ,

@mgdelacroix is OOO until the next year. Sorry for the delay in review.

commands/config_test.go
```diff
@@ -0,0 +1,105 @@
...
+			},
+		}
+
+		mockErrorMessage := "Mock Internal Server Error"
+		mockError := &model.AppError{Message: mockErrorMessage}
+
+		s.client.
+			EXPECT().
+			GetConfig().
+			Return(mockConfig, &model.Response{Error: mockError}).
```
agnivade: Let's return nil instead of `mockConfig` as I think that mocks the API more closely.

artemigkh:
I think someone already wrote tests for this and merged it, don't think this PR is necessary anymore

agnivade:
That seems right. @hanzei - should this be closed ?

hanzei:
That is correct @agnivade. Sorry for the inconvenient @artemigkh and thank you for the time you spend on the PR.

PR #73, opened 2019-12-07
fix model import on licence test file for testing

### LINKED ISSUE #70: test failes because of deprecated import
https://github.com/mattermost/mmctl/blob/954ce075d2a98fdc82a69ed01e5127b1d5de078b/commands/license_test.go#L4
This helps to run tests "github.com/mattermost/mattermost-server/v5/model"
is this a bug?

nadalfederer:
closes #70

### DIFF
--- a/commands/license_test.go
+++ b/commands/license_test.go
@@ -1,7 +1,7 @@
 package commands
 
 import (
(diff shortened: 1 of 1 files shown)


hanzei:
Good catch :+1:

PR #74, opened 2019-12-07
GH-13291 Add unit tests to the ChannelGroupEnableCmd mmctl command

lurcio:
**Summary**

Add unit tests to the "ChannelGroupEnableCmd" mmctl command.

**Ticket Link**

Fixes https://github.com/mattermost/mattermost-server/issues/13291

### DIFF
--- a/commands/group_test.go
+++ b/commands/group_test.go
@@ -1,0 +1,152 @@
+package commands
+
+import (
+	"strings"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mmctl/printer"
+	"github.com/spf13/cobra"
+)
+
+func (s *MmctlUnitTestSuite) TestChannelGroupEnableCmdF() {
+	s.Run("Enable group constrains with existing team and channel", func() {
+		teamArg := "team-id"
+		mockTeam := model.Team{Id: teamArg}
+		channelPart := "channel-id"
+		mockChannel := model.Channel{Id: channelPart}
+		channelArg := strings.Join([]string{teamArg, channelPart}, ":")
+		group := &model.Group{Name: "group-name"}
+		mockGroups := []*model.Group{group}
+		groupOpts := &model.GroupSearchOpts{
+			PageOpts: &model.PageOpts{
+				Page:    0,
+				PerPage: 10,
+			},
+		}
+
+		s.client.
+			EXPECT().
+			GetTeam(teamArg, "").
+			Return(&mockTeam, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			GetChannelByNameIncludeDeleted(channelPart, teamArg, "").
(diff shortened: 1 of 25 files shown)


commands/group_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		s.client.
+			EXPECT().
+			GetChannelByNameIncludeDeleted(channelPart, teamArg, "").
+			Return(&mockChannel, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			GetGroupsByChannel(channelPart, *groupOpts).
```
lurcio: Should I also test for when functions like this return errors?

mgdelacroix: Yes, it would be great if we had some tests checking the error cases for those calls :)

commands/group_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/spf13/cobra"
+)
+
+func (s *MmctlUnitTestSuite) TestChannelGroupEnableCmdF() {
+	s.Run("Enable group constrains with existing team and channel", func() {
+		teamArg := "team-id"
+		mockTeam := model.Team{Id: teamArg}
+		channelPart := "channel-id"
+		mockChannel := model.Channel{Id: channelPart}
+		channelArg := strings.Join([]string{teamArg, channelPart}, ":")
```
mgdelacroix: nit: this could be simplified with:

```suggestion
		channelArg := teamArg + ":" + channelPart
```

Same for the rest of instances

commands/group_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"strings"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mmctl/printer"
+	"github.com/spf13/cobra"
+)
+
+func (s *MmctlUnitTestSuite) TestChannelGroupEnableCmdF() {
+	s.Run("Enable group constrains with existing team and channel", func() {
+		teamArg := "team-id"
```
mgdelacroix: A `printer.Clean()` is needed at the beginning of the test case

commands/group_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Times(1)
+
+		err := channelGroupEnableCmdF(s.client, &cobra.Command{}, []string{channelArg})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+
+	s.Run("Enable group constrains with no associated groups", func() {
+		teamArg := "team-id"
```
mgdelacroix: Same `printer.Clean()` need. Even if the test case doesn't use it, it is better to keep it consistent between test cases

commands/group_test.go
lurcio: Thank you for your feedback, I'll add some tests for these calls

PR #88, opened 2019-12-24
MM-20752 - MMCTL config get option doesn't allow to pass plugin ids

reflog:
#### Summary
Added support for map keys with dots in their name. Allows get/set operations on paths like: 
`mmctl config set PluginSettings.PluginStates.com.mattermost.nps.Enable false`
`mmctl config get PluginSettings.PluginStates.com.mattermost.nps.Enable`
`mmctl config get PluginSettings.PluginStates.com.mattermost.nps`

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-20752

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -68,9 +68,17 @@ func init() {
 
 func getValue(path []string, obj interface{}) (interface{}, bool) {
 	r := reflect.ValueOf(obj)
-	val := r.FieldByName(path[0])
+	var val reflect.Value
+	if r.Kind() == reflect.Map {
+		val = r.MapIndex(reflect.ValueOf(path[0]))
+		if val.IsValid() {
+			val = val.Elem()
+		}
+	} else {
+		val = r.FieldByName(path[0])
+	}
 
-	if val.Kind() == reflect.Invalid {
+	if !val.IsValid() {
 		return nil, false
 	}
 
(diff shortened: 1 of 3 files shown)


hanzei:
cc @iomodo as you worked on this on the server CLI

commands/config_test.go
```diff
@@ -143,6 +143,49 @@ func (s *MmctlUnitTestSuite) TestConfigGetCmd() {
 		s.Require().Len(printer.GetLines(), 0)
 		s.Require().Len(printer.GetErrorLines(), 0)
 	})
+
+	s.Run("Get value if the key points to a map element", func() {
```
ethervoid: Please add more tests here:
- For the `PlugginSettings.Plugins` case with: single value, slice or even another map

commands/config_test.go
```diff
@@ -303,6 +346,58 @@ func (s *MmctlUnitTestSuite) TestConfigSetCmd() {
 		s.Require().Len(printer.GetLines(), 0)
 		s.Require().Len(printer.GetErrorLines(), 0)
 	})
+
+	s.Run("Set a field inside a map", func() {
+		printer.Clean()
```
ethervoid: Same here

commands/config.go
```diff
@@ -132,9 +155,33 @@ func setValue(path []string, obj reflect.Value, newValue interface{}) error {
...
-		return errors.New("Path object type is not supported")
+	} else if len(path) > 1 {
+		if val.Kind() == reflect.Struct {
+			return setValue(path[1:], val, newValue)
+		} else if val.Kind() == reflect.Map {
+
+			for i := len(path); i >= 1; i-- {
+				remainingPath := strings.Join(path[1:i], ".")
+
+				mapIter := val.MapRange()
```
ethervoid: Awesome. Didn't find it in my previous work :heart:

commands/config.go
```diff
@@ -78,6 +78,31 @@ func getValue(path []string, obj interface{}) (interface{}, bool) {
...
 		return getValue(path[1:], val.Interface())
+	} else if val.Kind() == reflect.Map {
+
+		for i := len(path); i >= 1; i-- {
+			remainingPath := strings.Join(path[1:i], ".")
+
+			mapIter := val.MapRange()
+
+			for mapIter.Next() {
+				if mapIter.Key().String() == remainingPath {
```
ethervoid: Why don't use [HasPrefix](https://golang.org/pkg/strings/#HasPrefix) to simplify?

commands/config.go
```diff
@@ -78,6 +78,31 @@ func getValue(path []string, obj interface{}) (interface{}, bool) {
 		return val.Interface(), true
 	} else if val.Kind() == reflect.Struct {
```
iomodo: `else` is not needed

commands/config.go
```diff
@@ -78,6 +78,31 @@ func getValue(path []string, obj interface{}) (interface{}, bool) {
...
 		return getValue(path[1:], val.Interface())
+	} else if val.Kind() == reflect.Map {
```
iomodo: `else` not needed

commands/config.go
```diff
@@ -132,9 +155,33 @@ func setValue(path []string, obj reflect.Value, newValue interface{}) error {
 				return errors.New("Target value type is not supported")
 			}
 		}
-	} else {
-		return errors.New("Path object type is not supported")
+	} else if len(path) > 1 {
```
iomodo: no `else` needed. I'd get rid of this `if` altogether, like: 
`if len(path)==0{
    return errors.New("Path object type is not supported")
}`

commands/config.go
reflog: `mmctl config get PluginSettings.Plugins.com.mattermost.demo-plugin.channelname` will not panic, but will return:
`Error: Invalid key`

commands/config.go
reflog: @iomodo  Here I believe it's needed, since it could also be other types like slice which we don't support yet

commands/config.go
reflog: why do you think that else is not needed here?

iomodo: It should cause error here `	val := r.FieldByName(path[0])`, because you have `map[string]map[string]interface{}` not `map[string]struct` and it has no field.

commands/config.go
iomodo: It's just "go way" :
```golang
if x = y1 {
    do stuff
    return ...
}
if x = y2 {
    do stuff
    return ...
}
```

If you have `return`-s there is no need of `else`.

PR #91, opened 2019-12-27
[GH13465] Add unit tests to the "ListTeamsCmd" mmctl command

larkox:
#### Summary

Add unit tests to the "ListTeamsCmd" mmctl command
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/13465

### DIFF
--- a/commands/team_test.go
+++ b/commands/team_test.go
@@ -633,8 +633,6 @@ func (s *MmctlUnitTestSuite) TestListTeamsCmdF() {
 	})
 
 	s.Run("One archived team", func() {
-		printer.Clean()
-
 		mockTeam := model.Team{
 			Name:     "Team1",
 			DeleteAt: 1,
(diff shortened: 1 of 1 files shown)


larkox:
My reasoning to include the test in this one and not on other tests is that I felt like there was no difference whether the channel was archived or not, and was not being shown to the user whether it was archived or not.

After a deeper thought, it is implicit in the json information (in json you can see the delete_at field with a value over 0), but even though, I wanted to make sure I properly tested both branches.

That being said, if we are offering both representations, shouldn't we be testing both?

agnivade:
> That being said, if we are offering both representations, shouldn't we be testing both?

Sure, we should. But then we should be doing it for all tests as well. But since you have already spent the effort in writing the tests, I don't see any harm in including them. I would just suggest to make the json and plain test variations to be sub-tests of the parent test, since both of them use the same input data. That would reduce some unnecessary boilerplate code.

agnivade:
I apologize for the confusion, but I meant "sub-tests", which would be another level of "s.Run()" inside the parent test.

larkox:
@agnivade Done! Thank you for the feedback!

PR #93, opened 2019-12-28
GH-13533 : Change channel rename command to be uniform with team rename

M-ZubairAhmed:
#### Summary
This PR updates the existing `Channel Rename` command to follow similar structure as `Team Rename`

#### Ticket Link
Fixes : https://github.com/mattermost/mattermost-server/issues/13533
Jira : None

#### Tests
Tests are intentionally not included so as to be taken at https://github.com/mattermost/mattermost-server/issues/13462 issue


agnivade:
Apologies for the delay on this @M-ZubairAhmed. I will get to this tomorrow.

commands/channel.go
```diff
@@ -23,12 +23,16 @@ var ChannelCreateCmd = &cobra.Command{
 	RunE: withClient(createChannelCmdF),
 }
 
+// ChannelRenameCmd is the command in mmctl to change name and/or display name of existing channel.
```
agnivade: I would suggest: "ChannelRenameCmd is used to change name and/or display name of an existing channel."

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
 }
 
 func renameChannelCmdF(c client.Client, cmd *cobra.Command, args []string) error {
-	var newDisplayName, newChannelName string
+	existingTeamChannel := args[0]
+	newChannelName, _ := cmd.Flags().GetString("name")
```
agnivade: I see that rename team also ignores the errors. Do you know why ?

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
+	newDisplayName, _ := cmd.Flags().GetString("display_name")
 
-	if len(args) < 2 {
-		return errors.New("Not enough arguments.")
+	// Both Display and name flags are not entered
+	if newDisplayName == "" && newChannelName == "" {
+		return errors.New("Require atleast one flag to rename team, either 'name' or 'display_name'")
```
agnivade: Copy paste error: rename _channel_ :)

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
 
 func renameChannelCmdF(c client.Client, cmd *cobra.Command, args []string) error {
-	var newDisplayName, newChannelName string
+	existingTeamChannel := args[0]
+	newChannelName, _ := cmd.Flags().GetString("name")
+	newDisplayName, _ := cmd.Flags().GetString("display_name")
 
-	if len(args) < 2 {
-		return errors.New("Not enough arguments.")
+	// Both Display and name flags are not entered
```
agnivade: I would suggest: "Atleast one of the flags must be present."

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
-		channelPatch.DisplayName = &newDisplayName
+		patchedChannel.DisplayName = &newDisplayName
 	}
 
-	if _, response := c.PatchChannel(channel.Id, &channelPatch); response.Error != nil {
-		return response.Error
+	// Using PatchChannel API to rename channel
+	_, response := c.PatchChannel(channel.Id, patchedChannel)
+	if response.Error != nil {
+		return errors.New("Cannot rename team '" + existingTeamChannel + "', error : " + response.Error.Error())
```
agnivade: Same: rename _channel_. Also `existingTeamChannel` contains both team and channel names, so cannot rename channel is not really accurate there. Let's pass the `channel.Name` here instead of existingTeamChannel.

docs/mmctl_user_email.md
```diff
@@ -13,8 +13,7 @@ mmctl user email [user] [new email] [flags]
 ### Examples
 
 ```
-  user email test user@example.com
-  user activate username
```
agnivade: This seems to be unrelated.

docs/mmctl_channel_rename.md
```diff
@@ -4,23 +4,26 @@ Rename a channel
...
+  channel rename myteam:oldchannel --name 'new-channel'
+  channel rename myteam:oldchannel --display_name 'New Display Name'
 ```
 
 ### Options
 
 ```
       --display_name string   Channel Display Name
   -h, --help                  help for rename
+      --name string           Channel Name
```
agnivade: Let's move this above help.

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
 	}
 
+	printer.Print("'" + existingTeamChannel + "' channel renamed")
```
agnivade: Same as above: let's use `channel.Name`

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
+	// Both Display and name flags are not entered
+	if newDisplayName == "" && newChannelName == "" {
+		return errors.New("Require atleast one flag to rename team, either 'name' or 'display_name'")
 	}
 
-	channel := getChannelFromChannelArg(c, args[0])
+	channel := getChannelFromChannelArg(c, existingTeamChannel)
 	if channel == nil {
-		return errors.New("Unable to find channel '" + args[0] + "'")
+		return errors.New("Unable to find channel '" + existingTeamChannel + "'")
```
agnivade: I would suggest a slight modification to be more accurate: "Unable to find channel from ...". Because `existingTeamChannel` is not just the channel name.

commands/channel.go
```diff
@@ -345,32 +350,35 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
 	}
 
-	newChannelName = args[1]
-	newDisplayName, errdn := cmd.Flags().GetString("display_name")
-	if errdn != nil {
-		return errdn
+	patchedChannel := &model.ChannelPatch{}
```
agnivade: I thought `channelPatch` as the variable name was better. It's not really a "patched" channel because the patch is going to happen and this is the patch. Can you explain your reasoning behind selecting this name ?

commands/channel.go
M-ZubairAhmed: I am not able to find where but i think @mgdelacroix suggested that if there was an error in getting flag, then the flag variable would be empty and error will be thrown in the next code step. Anyways what do you suggest

docs/mmctl_user_email.md
M-ZubairAhmed: this is because of `make docs`

docs/mmctl_channel_rename.md
M-ZubairAhmed: this is autogenerated, i didnt do it manually

docs/mmctl_user_email.md
agnivade: Ah okay.

commands/channel.go
agnivade: I think that might clash with explicitly passing an empty value. In this case, we need a value, so this is fine. But in some cases, we might have a use case for passing empty value. In that case, that check is necessary. And I do see the error being checked in some parts of the codebase, so for the sake of consistency I suggest to add the check.

M-ZubairAhmed: got it

M-ZubairAhmed: i will add now

M-ZubairAhmed: But i also see in many places it is being checked to see if flag is being empty. What error message should it be here in our case ?

commands/channel.go
```diff
@@ -345,32 +350,43 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
 
 func renameChannelCmdF(c client.Client, cmd *cobra.Command, args []string) error {
-	var newDisplayName, newChannelName string
+	existingTeamChannel := args[0]
 
-	if len(args) < 2 {
-		return errors.New("Not enough arguments.")
+	newChannelName, err := cmd.Flags().GetString("name")
+	if err != nil {
+		return errors.New("Cannot get name flag, error : " + err.Error())
```
agnivade: Let's just directly return the error. It will be of the form: `fmt.Errorf("flag accessed but not defined: %s", name)`. So the name is already there.

commands/channel.go
```diff
@@ -345,32 +350,43 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
 	}
 
-	channel := getChannelFromChannelArg(c, args[0])
-	if channel == nil {
-		return errors.New("Unable to find channel '" + args[0] + "'")
+	newDisplayName, err := cmd.Flags().GetString("display_name")
+	if err != nil {
+		return errors.New("Cannot get display_name flag, error : " + err.Error())
```
agnivade: Same as above.

docs/mmctl_channel_rename.md
agnivade: Oh it's alphabetical. Got it.

commands/channel.go
```diff
@@ -345,32 +350,43 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
+		return err
 	}
 
-	newChannelName = args[1]
-	newDisplayName, errdn := cmd.Flags().GetString("display_name")
-	if errdn != nil {
-		return errdn
+	// Atleast one of display name or name flag must be present
+	if newDisplayName == "" && newChannelName == "" {
+		return errors.New("Require atleast one flag to rename channel, either 'name' or 'display_name'")
```
mgdelacroix: Typo:

```suggestion
		return errors.New("Require at least one flag to rename channel, either 'name' or 'display_name'")
```

commands/channel.go
```diff
@@ -345,32 +350,43 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
+	newDisplayName, err := cmd.Flags().GetString("display_name")
+	if err != nil {
+		return err
 	}
 
-	newChannelName = args[1]
-	newDisplayName, errdn := cmd.Flags().GetString("display_name")
-	if errdn != nil {
-		return errdn
+	// Atleast one of display name or name flag must be present
```
mgdelacroix: Typo:

```suggestion
	// At least one of display name or name flag must be present
```

commands/channel.go
```diff
@@ -345,32 +350,43 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
...
 
-	if _, response := c.PatchChannel(channel.Id, &channelPatch); response.Error != nil {
-		return response.Error
+	// Using PatchChannel API to rename channel
+	_, response := c.PatchChannel(channel.Id, channelPatch)
+	if response.Error != nil {
+		return errors.New("Cannot rename channel '" + channel.Name + "', error : " + response.Error.Error())
 	}
 
+	printer.Print("'" + channel.Name + "' channel renamed")
```
mgdelacroix: As this is an update action, we should print the updated channel in case the user requests the JSON format. So if we modify the patch line to get the updated channel:

```go
updatedChannel, response := c.PatchChannel(channel.Id, channelPatch)
```

We can modify the print to be:

```suggestion
	printer.PrintT("'{{.Name}}' channel renamed", updatedChannel)
```

PR #94, opened 2019-12-28
GH-13460: Add unit tests for AddChannelUsersCmd mmctl command.

tasdomas:
#### Summary
Unit tests for the AddChannelUsersCmd have been added.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/13460

### DIFF
--- a/commands/channel_test.go
+++ b/commands/channel_test.go
@@ -12,7 +12,6 @@ import (
 
 func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
 	s.Run("Search for an existing channel on an existing team", func() {
-		defer printer.Clean()
 		teamArg := "example-team-id"
 		mockTeam := model.Team{Id: teamArg}
 		channelArg := "example-channel"
@@ -41,7 +40,7 @@ func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
 	})
 
 	s.Run("Search for an existing channel without specifying team", func() {
-		defer printer.Clean()
+		printer.Clean()
 		teamId := "example-team-id"
 		otherTeamId := "example-team-id-2"
 		mockTeams := []*model.Team{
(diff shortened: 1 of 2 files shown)


commands/channel_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
 	s.Run("Search for an existing channel on an existing team", func() {
+		defer printer.Clean()
```
mgdelacroix: Although the clean could be done at the end of the function's execution, we should keep it without the `defer` for consistency

tasdomas: Sure. Fixed.

tasdomas:
@mgdelacroix - there already was a test adding both an existing and a nonexistent user, it was just not titled clearly. I've updated that.

commands/channel_test.go
mgdelacroix: Tests under `TestAddChannelUsersCmdF` don't have the `defer` anymore, but the `printer.Clean()` call should be moved to be the first thing at the beginning of each test case.

Tests under `TestSearchChannelCmdF` still have the `defer` keyword.

tasdomas: @mgdelacroix - I've updated the tests.

Though I have to say that with the way the printer cleanup is currently done (at the start of every subtest that is not the first subtest to run), it makes the subtests depend on the order they're arranged in. If somebody were to move the first subtest down, it would start failing,

And conceptually, if the `printer.Clean()` statement is at the start of the subtest, that means the test is cleaning up the output of another subtest. Whereas a `printer.Clean()` at the end of the subtest means it's cleaning up its own output - it feels more contained that way. Though we do need a `defer` in that case.

mgdelacroix: @tasdomas thanks for the changes. Please bear in mind that the idea is to have the `printer.Clean()` call at the beginning of every test case, including the first one, to avoid the order problem, as noted in [the campaign blog post](https://developers.mattermost.com/blog/unit-testing-mmctl-commands/#printer-cleanup).

Having said that, I agree on the `defer printer.Clean()` approach being clearer conceptually speaking. To avoid breaking the consistency, I think that the best approach is to keep following the current rules, and when the campaign is finished, we can add all the `defer`s in one PR.

tasdomas: @mgdelacroix - sure, no problem. I've updated my branch to add `printer.Clean()` for first subtests too.

tasdomas: @mgdelacroix  - if there are any other changes you'd like to see added to this PR, let me know.

PR #95, opened 2019-12-30
[MM-21382] Add unit tests to the "ChannelCreateCmd" mmctl command

phillipahereza:
#### Summary

This PR covers the addition of unit tests to the ChannelCreateCmd command, in the `commands/channel.go` file.

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-21382


commands/channel_test.go
```diff
@@ -129,3 +129,150 @@ func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
...
+
+		cmd := &cobra.Command{}
+		cmd.Flags().String("name", mockChannelName, "")
+		err := createChannelCmdF(s.client, cmd, []string{})
+
+		s.Require().Equal(err.Error(),"Display Name is required")
+		s.Require().Len(printer.GetLines(), 0)
+	})
+
+	s.Run("Create channel with a Name but no Display Name returns an error", func() {
```
jespino: This description is wrong, here we are creating a channel with Displa Name bot no Name

commands/channel_test.go
```diff
@@ -129,3 +129,150 @@ func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
...
+			EXPECT().
+			CreateChannel(mockChannel).
+			Return(mockChannel, &model.Response{Error: nil}).
+			Times(1)
+
+		err := createChannelCmdF(s.client, cmd, []string{})
+		s.Require().Nil(err)
+		s.Require().Equal(mockChannel, printer.GetLines()[0])
+		s.Require().Len(printer.GetLines(), 1)
+	})
```
jespino: I miss here tests where the API calls for `GetTeam`fails, and one for the `CreateChannel` API call fails.

phillipahereza: @jespino by fail you mean return an error or when `GetTeam` returns a `nil`

jespino: I mean when GetTeam returns an error.

phillipahereza: @jespino I might be wrong but the no error is returned by `getTeamFromTeamArg` and if I change it to return the error from `GetTeam` won't I break other parts of the code that use `getTeamFromTeamArg`

jespino: Well... the underneath code is ignoring the error, so... yes, maybe makes no sense to test the error returned here. Actually I'm not very happy with how is handled the errors here, but that is another subject that we should discuss for other PR.

jespino: @mgdelacroix What do you thing, we are swallowing the error here, should we raise the error to the interface some how, or at least leave the commands the possibility of handling the error differently than not-found teams?

commands/channel_test.go
```diff
@@ -129,3 +129,150 @@ func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
...
+
+		cmd := &cobra.Command{}
+		cmd.Flags().String("display_name", mockChannelDisplayName, "")
+		err := createChannelCmdF(s.client, cmd, []string{})
+
+		s.Require().Equal(err.Error(),"Name is required")
+		s.Require().Len(printer.GetLines(), 0)
+	})
+
+	s.Run("Create channel with a Name and Display Name but Team returns an error", func() {
```
mgdelacroix:
```suggestion
	s.Run("Create channel with a Name and Display Name but no Team returns an error", func() {
```

commands/channel_test.go
```diff
@@ -129,3 +129,150 @@ func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
...
+		s.client.
+			EXPECT().
+			CreateChannel(mockChannel).
+			Return(mockChannel, &model.Response{Error: nil}).
+			Times(1)
+
+		err := createChannelCmdF(s.client, cmd, []string{})
+		s.Require().Nil(err)
+		s.Require().Equal(mockChannel, printer.GetLines()[0])
+		s.Require().Len(printer.GetLines(), 1)
```
mgdelacroix: nit: Length check should go before the equal tests that assumes the lines slice contains at least one

commands/channel_test.go
```diff
@@ -129,3 +129,150 @@ func (s *MmctlUnitTestSuite) TestSearchChannelCmdF() {
...
+	})
+
+	s.Run("Create invite only channel", func() {
```
mgdelacroix: For consistency with the naming, I think we should use `private` here:

```suggestion
	s.Run("Create private channel", func() {
```

commands/channel_test.go
mgdelacroix: Totally agree. I can see the reasoning on why this was done at the beginning (to ease the use of the `getTeamsFromTeamArgs`) but I think this can be done better. Same happens in almost every `From*Args` function, so this may very well be a good refactor

mgdelacroix: In any case, I'd say we should not tackle this change as part of the PR

jasonblais:
Hey @phillipahereza - touching base to see if you have any questions about the feedback above? We'd be happy to help clarify :)

hanzei:
This PR has become orphaned and will be closed. Thanks for your contribution @phillipahereza :+1:

PR #100, opened 2020-01-08
Ldflags

ethervoid:
#### Summary

This PR changes how we configure the build hash and the version to be able to pass it through environment variables so we can do it in the brew compilation

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-19068

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
 .PHONY: vendor docs mocks
 GO_PACKAGES=$(shell go list ./...)
 GO ?= $(shell command -v go 2> /dev/null)
-BUILD_HASH = $(shell git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -48,9 +52,11 @@ gofmt:
 
 govet:
 	@echo Running govet
-	env GO111MODULE=off $(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
 	$(GO) vet $(GO_PACKAGES)
-	$(GO) vet -vettool=$(GOPATH)/bin/shadow $(GO_PACKAGES)
+       ifeq ($(ADVANCED_VET), TRUE)
```
ethervoid: Space vs tabs and here spaces are needed :(

PR #104, opened 2020-01-15
[MM-21705] Make sure docs are updated

RajatVaryani:
#### Summary
This PR adds a check in lint stage to make sure the docs are updated. The stage fails when the docs are not updated. 
Fixes https://github.com/mattermost/mattermost-server/issues/13625

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,6 +148,7 @@ workflows:
   version: 2
   untagged-build:
     jobs:
+      - docs
       - lint
       - setup
       - test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -13,6 +13,9 @@ jobs:
       - checkout
       - run:
           command: |
+            echo "Making sure docs are updated"
+            make docs
+            [[ -z $(git status -s) ]] || exit 1
```
ethervoid: Sending an error message saying that the docs are outdated will be nice

RajatVaryani: Done.

RajatVaryani:
> Thanks for the contribution @RajatVaryani!!
> 
> Logic looks good, would you be up to changing it to run on an independent `docs` job?

@mgdelacroix I have made the changes.

PR #107, opened 2020-01-20
Start using golanci linter

ethervoid:
#### Summary

This PR changes the current govet configuration to start using golanci as our linter. It also fixes all the errors the lintern is reporting

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21706


.golangci.yml
```diff
@@ -0,0 +1,43 @@
...
+    - varcheck
+    - errcheck
+
+issues:
+  exclude-rules:
+    - linters:
+      # ignore unused warnings from enterprise code
+      # add more as required.
+      - unused
+      text: "testLogsCmdF|MmctlE2ETestSuite|MmctlUnitTestSuite|setupTestHelper"
```
ethervoid: Estos los marca como  `unused` y supongo que es por que son utilizados en los tests

.golangci.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - linters:
+      # ignore golint error for a lot of packages for now
+      - golint
+      path: "client"
```
ethervoid: To keep consistency with the server-side functions. In the future we can change those too

commands/command.go
```diff
@@ -58,11 +58,11 @@ func init() {
 	CommandCreateCmd.Flags().String("title", "", "Command Title")
 	CommandCreateCmd.Flags().String("description", "", "Command Description")
 	CommandCreateCmd.Flags().String("trigger-word", "", "Command Trigger Word (required)")
-	CommandCreateCmd.MarkFlagRequired("trigger-word")
+	_ = CommandCreateCmd.MarkFlagRequired("trigger-word")
```
ethervoid: I suppose we don't need to act in case of error that's why I discard these errors

.circleci/config.yml
```diff
@@ -22,12 +22,15 @@ jobs:
       name: default
     steps:
       - checkout
+      - attach_workspace:
+          at: ~/go/mattermost/
       - run:
+          name: "Validate lint"
           command: |
+            echo "installing golangci-lint"
+            curl -sfl https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0 # keep this in sync with .golangci.yml
```
hanzei: The Plugin Marketplace also uses golangci with circleci. You might want to just copy and past the command from there: https://github.com/mattermost/mattermost-marketplace/pull/27/files#diff-1d37e48f9ceff6d8030570cd36286a61

The advantage of this implementation is, that the binary gets cached, which speeds up the build.

.golangci.yml
```diff
@@ -0,0 +1,43 @@
+service:
```
hanzei: You might want to use a more aggressive set of linter similar to the ones in https://github.com/mattermost/mattermost-marketplace/pull/27/files#diff-e23b7926f022c85af1665b36d256231b

Just as an idea as these linter have been proven to be valuable.

.circleci/config.yml
ethervoid: Oh good point. I've copied from our service circleci file, but this one is better. Thank you

.golangci.yml
ethervoid: Copied all the linters used

ethervoid:
@hanzei re-review? :D

commands/docs.go
```diff
@@ -28,17 +27,20 @@ func init() {
 func docsCmdF(cmd *cobra.Command, args []string) error {
 	outDir, _ := cmd.Flags().GetString("directory")
 
-	if fileInfo, err := os.Stat(outDir); os.IsNotExist(err) {
+	fileInfo, err := os.Stat(outDir)
+
+	switch {
```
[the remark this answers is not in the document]
ethervoid: Because of the `gocritic` linter:

```
commands/docs.go:32:2: ifElseChain: rewrite if-else to switch statement (gocritic)
	if os.IsNotExist(err) {
```

commands/user_e2e_test.go
```diff
@@ -5,6 +5,7 @@ package commands
 
 import (
 	"github.com/mattermost/mattermost-server/v5/model"
+
 	"github.com/mattermost/mmctl/printer"
 
 	"github.com/spf13/cobra"
```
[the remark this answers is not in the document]
ethervoid: That is done by the `gofmt -ed` required by the `goimports` linter:
```
File is not `goimports`-ed with -local github.com/mattermost/mmctl (goimports)go-lint
```

printer/printer.go
```diff
@@ -43,14 +43,14 @@ func SetSingle(single bool) {
 // template
 func PrintT(templateString string, v interface{}) {
 	switch printer.Format {
-	case FORMAT_PLAIN:
+	case FormatPlain:
 		t := template.Must(template.New("").Parse(templateString))
 		var tpl bytes.Buffer
-		t.Execute(&tpl, v)
+		_ = t.Execute(&tpl, v)
```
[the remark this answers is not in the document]
ethervoid: Yes, I was thinking what to do here and is hard to decide what is the best way so I left that way by now. 

Giving it a second though could be useful to catch that error and call the `PrintError` in case we have an error parsing the print template. What do you think?

commands/completion.go
```diff
@@ -48,7 +48,7 @@ func init() {
 }
 
 func bashCmdF(cmd *cobra.Command, args []string) {
-	RootCmd.GenBashCompletion(os.Stdout)
+	_ = RootCmd.GenBashCompletion(os.Stdout)
```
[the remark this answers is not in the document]
ethervoid: Ooops, good catch. Missed that one thinking that it was like other Cobra command initialization. Fixed

commands/completion.go
```diff
@@ -48,7 +48,10 @@ func init() {
 }
 
 func bashCmdF(cmd *cobra.Command, args []string) {
```
hanzei: Nit: instead of logging the error you can also return it and let cobra command print it
```suggestion
func bashCmdF(cmd *cobra.Command, args []string) error {
```

The you also would have up update https://github.com/mattermost/mmctl/pull/107/files#diff-0f095041bcad19744b6226478e23d155R38 to 
```golang
	RunE: zshCmdF,
```

This is more of an FYI and not blocking for the PR.

commands/user_e2e_test.go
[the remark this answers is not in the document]
hanzei: I think the correct order should be:

```go
	"github.com/mattermost/mattermost-server/v5/model"
	"github.com/spf13/cobra"

	"github.com/mattermost/mmctl/printer"
```

commands/docs.go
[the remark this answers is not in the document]
hanzei: The complain for `gocritic` is about using `if-else`. That is correct the old code should not have used `if-else`. My proposal would be to just space the code out a bit:
```go
fileInfo, err := os.Stat(outDir)
if err != nil {
	if !os.IsNotExist(err) {
		return err
	}
	
	err = os.Mkdir(outDir, 0755)
	if err != nil {
		return err
	}
}

if !fileInfo.IsDir() {
	return errors.New(fmt.Sprintf("File \"%s\" is not a directory", outDir))
}
```

At the end this is mostly a subjective opinion. Feel free to ignore my comment.

printer/printer.go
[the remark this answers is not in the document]
hanzei: :+1: for just calling `PrintError`.

ethervoid:
Thank you @mgdelacroix and @hanzei  :heart:

PR #111, opened 2020-01-26
[MM-21921] Adds the server version check before running a command

mgdelacroix:
#### Summary
Before running any command that uses the Client instance, `mmctl` will
now check that the server version matches its own, and will show an
error if it doesn't.

This PR enables as well the `--strict` flag and the `MMCTL_STRICT`
environment variable; if one of those is set, `mmctl` will error in case
the server version doesn't match and will not run the command.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21921


commands/init.go
```diff
@@ -4,58 +4,80 @@
...
 	return func(cmd *cobra.Command, args []string) error {
-		c, err := InitClient()
+		c, serverVersion, err := InitClient()
 		if err != nil {
 			return err
 		}
+
+		valid := CheckVersionMatch(Version, serverVersion)
+		if !valid {
+			if viper.GetBool("strict") {
```
ethervoid: One question to understand better. Do we need viper as a new dependency?

I mean, we can just create a new method in `version.go` called something like `SetStrict` like we have with the `printer.SetFormat` and then use it here as we used `Version` variable in the line 32

Makefile
```diff
@@ -7,7 +7,6 @@ BUILD_VERSION ?= $(shell git ls-remote --tags --refs --sort="v:refname" git://gi
 ADVANCED_VET ?= TRUE
 
 LDFLAGS += -X "github.com/mattermost/mmctl/commands.BuildHash=$(BUILD_HASH)"
-LDFLAGS += -X "github.com/mattermost/mmctl/commands.Version=$(BUILD_VERSION)"
```
ethervoid: Just a comment, not a proposal. One thing that worries my about having version variables is that they tend to keep outdated.

commands/init_test.go
```diff
@@ -0,0 +1,64 @@
...
+		{
+			Name: "Minor version is less",
+			Version: "1.2.3",
+			ServerVersion: "1.1.3",
+			Expected: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
```
ethervoid: :heart:

commands/init.go
mgdelacroix: I'm not sure I'm understanding your point here. Viper is used so when reading the `strict` value, it takes into consideration both the `--strict` flag and the `MMCTL_STRICT` environment variable. As a side benefit, I've binded as well the `MMCTL_FORMAT` environment variable, so now both can be received as flags or env variables.

The rationale behind this is that we want to allow the admins using `mmctl` that want all their commands to be strict to be able to do it easily, and an environment variable is a great mechanism for this.

ethervoid: Cool, understood! Thank you

Makefile
mgdelacroix: I think that the version variable allows us to keep a different version number for each branch (say, if we have to release a 5.20 and a 5.19.1 versions of the tool), but it is true that we have to be careful for it not to become outdated.

PR #117, opened 2020-01-30
[MM-21921] Adds the server version check before running a command

mgdelacroix:
#### Summary
v5.20 cherry pick for https://github.com/mattermost/mmctl/pull/111

It has the version changed to `5.20` in the `commands/version.go` file from the original PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21921

PR #118, opened 2020-01-30
Fixes the error message shown when the strict mode is enabled

mgdelacroix:
#### Summary
Old message was not updated correctly. This PR also makes the command exit with an error instead of printing the help message

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -4,7 +4,7 @@
 package commands
 
 import (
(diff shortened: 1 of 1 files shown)


mgdelacroix:
This cherry-pick needs to be done after https://github.com/mattermost/mmctl/pull/117 is merged. Will do manually

PR #121, opened 2020-02-07
[MM-22166] Rejects SHA1 intermediate certificates and allows for a bypass

mgdelacroix:
#### Summary
The client will now reject to communicate with a server that has a SHA1 certificate in the chain (in every position but the root one). It will allow as well to use a `--insecure-sha1-intermediate` flag and a `MMCTL_INSECURE_SHA1_INTERMEDIATE` environment variable to bypass this check and connect to insecure servers.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22166

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,7 +48,6 @@ linters:
     - interfacer
     - misspell
     - nakedret
-    - scopelint
     - staticcheck
     - stylecheck
     - typecheck
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/mattermost/mattermost-server/v5/model"
@@ -135,6 +136,8 @@ func loginCmdF(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	allowInsecure := viper.GetBool("insecure-sha1-intermediate")
+
 	url := args[0]
 	method := MethodPassword
 
@@ -174,10 +177,10 @@ func loginCmdF(cmd *cobra.Command, args []string) error {
 		var c *model.Client4
 		var err error
 		if mfaToken != "" {
-			c, _, err = InitClientWithMFA(username, password, mfaToken, url)
+			c, _, err = InitClientWithMFA(username, password, mfaToken, url, allowInsecure)
 			method = MethodMFA
 		} else {
-			c, _, err = InitClientWithUsernameAndPassword(username, password, url)
+			c, _, err = InitClientWithUsernameAndPassword(username, password, url, allowInsecure)
 		}
 		if err != nil {
 			printer.PrintError(err.Error())
@@ -192,7 +195,7 @@ func loginCmdF(cmd *cobra.Command, args []string) error {
 			InstanceURL: url,
 			AuthToken:   accessToken,
 		}
-		if _, _, err := InitClientWithCredentials(&credentials); err != nil {
+		if _, _, err := InitClientWithCredentials(&credentials, allowInsecure); err != nil {
 			printer.PrintError(err.Error())
 			// We don't want usage to be printed as the command was correctly built
 			return nil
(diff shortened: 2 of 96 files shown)


commands/init.go
```diff
@@ -44,26 +58,58 @@ func withClient(fn func(c client.Client, cmd *cobra.Command, args []string) erro
...
+				VerifyPeerCertificate: func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+					// loop over certificate chains
+					for _, chain := range verifiedChains {
+						// check all certs but the root one
+						certs := chain[:len(chain)-1]
+
+						for _, cert := range certs {
+							for _, alg := range insecureSignatureAlgorithms {
+								if cert.SignatureAlgorithm == alg {
+									return fmt.Errorf("insecure algorithm %s found in the certificate chain. Use --insecure-sha1-intermediate flag to ignore. Aborting", alg)
```
jupenur: In case there are multiple chains, we only need one to be valid. This will fail if even a single one is invalid.

mgdelacroix: Updated the logic and added tests, PTAL @jupenur

.golangci.yml
```diff
@@ -48,7 +48,6 @@ linters:
     - interfacer
     - misspell
     - nakedret
-    - scopelint
```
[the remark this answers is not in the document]
mgdelacroix: :wink:

commands/init.go
```diff
@@ -44,26 +58,67 @@ func withClient(fn func(c client.Client, cmd *cobra.Command, args []string) erro
 	}
 }
 
-func InitClientWithUsernameAndPassword(username, password, instanceURL string) (*model.Client4, string, error) {
+func isValidChain(chain []*x509.Certificate) bool {
+	// check all certs but the root one
+	certs := chain[:len(chain)-1]
+
+	for _, cert := range certs {
+		for _, alg := range insecureSignatureAlgorithms {
```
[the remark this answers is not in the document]
mgdelacroix: Updated, PTAL :)

commands/auth.go
```diff
@@ -174,10 +177,10 @@ func loginCmdF(cmd *cobra.Command, args []string) error {
 		var c *model.Client4
 		var err error
 		if mfaToken != "" {
-			c, _, err = InitClientWithMFA(username, password, mfaToken, url)
+			c, _, err = InitClientWithMFA(username, password, mfaToken, url, allowInsecure)
```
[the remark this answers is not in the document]
mgdelacroix: Agree, but I think such a refactor would be better done apart from this PR.

commands/init.go
```diff
@@ -44,26 +58,65 @@ func withClient(fn func(c client.Client, cmd *cobra.Command, args []string) erro
...
+}
+
+func VerifyCertificates(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	// loop over certificate chains
+	for _, chain := range verifiedChains {
+		if isValidChain(chain) {
+			return nil
+		}
+	}
+	return fmt.Errorf("insecure algorithm found in the certificate chain. Use --insecure-sha1-intermediate flag to ignore. Aborting")
```
jespino: I would end here the sentence with something like `Use --insecure-sha1-intermediate flag to ignore under your own responsability`, what do you think @wiersgallak?

commands/init.go
```diff
@@ -44,26 +58,65 @@ func withClient(fn func(c client.Client, cmd *cobra.Command, args []string) erro
...
+
+	for _, cert := range certs {
+		if _, ok := insecureSignatureAlgorithms[cert.SignatureAlgorithm]; ok {
+			return false
+		}
+	}
+	return true
+}
+
+func VerifyCertificates(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
jespino: The rawCerts are received here, but not used.

commands/init_test.go
```diff
@@ -55,11 +56,67 @@ func TestCheckVersionMatch(t *testing.T) {
...
+					{SignatureAlgorithm: x509.SHA1WithRSA},
+				},
+				{
+					{SignatureAlgorithm: x509.SHA256WithRSA},
+					{SignatureAlgorithm: x509.SHA256WithRSA},
+					{SignatureAlgorithm: x509.SHA1WithRSA},
+				},
+			},
+			ExpectedError: false,
+		},
```
jespino: Maybe another test with 2 invalid chains?

commands/init.go
mgdelacroix: They are there to comply with the function signature that `tls.Config.VerifyPeerCertificate` expects

jespino: Makes sense. Thanks for the explanation.

PR #130, opened 2020-03-04
MM-20913 implement getbusy, setbusy, clearbusy

wiggin77:
#### Summary

This PR implements the following commands in mmctl:
- system getbusy
- system setbusy [seconds]
- system clearbusy

These APIs get/set/clear the busy state of the server, which disables/enables non-critical services when under high load.

Blocked by: https://github.com/mattermost/mattermost-server/pull/13986

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-20913


commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/mattermost/mmctl/client"
+	"github.com/mattermost/mmctl/printer"
+)
+
+var SystemCmd = &cobra.Command{
+	Use:   "system",
+	Short: "System management",
```
mgdelacroix: Can we add a `Long` description here?

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+var SystemGetBusyCmd = &cobra.Command{
+	Use:     "getbusy",
+	Short:   "Get the current busy state",
+	Long:    `Gets the server busy state (high load) and timestamp corresponding to when the server busy flag will be automatically cleared.`,
+	Example: `  system getbusy`,
```
mgdelacroix:
```suggestion
	Example: `  system getbusy`,
	Args:    cobra.NoArgs,
```

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+	Example: `  system setbusy 3600`,
+	Args:    cobra.ExactArgs(1),
+	RunE:    withClient(setBusyCmdF),
+}
+
+var SystemClearBusyCmd = &cobra.Command{
+	Use:     "clearbusy",
+	Short:   "Clears the busy state",
+	Long:    `Clear the busy state, which re-enables non-critical services.`,
+	Example: `  system clearbusy`,
```
mgdelacroix:
```suggestion
	Example: `  system clearbusy`,
	Args:    cobra.NoArgs,
```

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+	RunE:    withClient(clearBusyCmdF),
+}
+
+func init() {
+	SystemSetBusyCmd.Flags().Uint("seconds", 3600, "Number of seconds until server is automatically marked as not busy.")
```
mgdelacroix: We could use here the shortflag format as well:

```suggestion
	SystemSetBusyCmd.Flags().UintP("seconds", "s", 3600, "Number of seconds until server is automatically marked as not busy.")
```

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+	SystemCmd.AddCommand(
+		SystemGetBusyCmd,
+		SystemSetBusyCmd,
+		SystemClearBusyCmd,
+	)
+	RootCmd.AddCommand(SystemCmd)
+}
+
+func getBusyCmdF(c client.Client, cmd *cobra.Command, args []string) error {
```
mgdelacroix: I know this is inconsistently done in other commands, but if we are not using `args` we should use a blank assignment:

```suggestion
func getBusyCmdF(c client.Client, cmd *cobra.Command, _ []string) error {
```

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+	ok, response := c.SetServerBusy(int(seconds))
+	if response.Error != nil || !ok {
+		printer.PrintError(fmt.Sprintf("Unable to set busy state: %v", response.Error))
+		return response.Error
+	}
+	printer.Print("Busy state set")
+	return nil
+}
+
+func clearBusyCmdF(c client.Client, cmd *cobra.Command, args []string) error {
```
mgdelacroix: Same:

```suggestion
func clearBusyCmdF(c client.Client, cmd *cobra.Command, _ []string) error {
```

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+	ok, response := c.ClearServerBusy()
+	if response.Error != nil || !ok {
+		printer.PrintError(fmt.Sprintf("Unable to clear busy state: %v", response.Error))
```
mgdelacroix: There is the case of `response.Error == nil && ok == false` in which case we would be printing `nil` as part of the error message

commands/user_test.go
```diff
@@ -591,7 +591,7 @@ func (s *MmctlUnitTestSuite) TestUserInviteCmd() {
 			Return(resultTeamModels[3], &model.Response{Error: nil}).
 			Times(1)
 
-		// Setup InvitUsersToTeam
+		// Setup InviteUsersToTeam
```
mgdelacroix: Nice catch!

commands/system.go
mgdelacroix: We are setting here a flag for `seconds` but we are using a positional argument in the command itself. I think we should mark the flag as required, add the `Args: cobra.NoArgs` to the command and modify the logic to pick up seconds from the flag instead of `args[0]`

commands/system_test.go
```diff
@@ -0,0 +1,148 @@
...
+			Times(1)
+
+		err := setBusyCmdF(s.client, &cobra.Command{}, []string{})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0], map[string]string{"status": "ok"})
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+
+	s.Run("SetBusy invalid seconds", func() {
```
larkox: Probably corner case handled by ParseUInt, but what about a number > MaxInt?

commands/system_test.go
```diff
@@ -0,0 +1,148 @@
...
+			Times(1)
+
+		err := clearBusyCmdF(s.client, &cobra.Command{}, []string{})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0], map[string]string{"status": "ok"})
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+
+	s.Run("ClearBusy with error", func() {
```
larkox: On top of @mgdelacroix 's comment, you should add a test for error = nil, ok = false

commands/system_test.go
```diff
@@ -0,0 +1,148 @@
...
+		s.client.
+			EXPECT().
+			GetServerBusy().
+			Return(nil, &model.Response{Error: &model.AppError{Id: "Mock Error"}}).
+			Times(1)
+
+		err := getBusyCmdF(s.client, &cobra.Command{}, []string{})
+		s.Require().NotNil(err)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 1)
```
larkox: Maybe I am nitpicking, but wouldn't it be nice to make sure the error you are receiving is the error you are expecting? Not sure if this will make the test too brittle, but may make sense.

commands/system_test.go
```diff
@@ -0,0 +1,148 @@
...
+	s.Run("GetBusy with error", func() {
+		printer.Clean()
+		s.client.
+			EXPECT().
+			GetServerBusy().
+			Return(nil, &model.Response{Error: &model.AppError{Id: "Mock Error"}}).
+			Times(1)
+
+		err := getBusyCmdF(s.client, &cobra.Command{}, []string{})
+		s.Require().NotNil(err)
```
larkox: I think we are preferring Require().Error(err) for these cases.

commands/system_test.go
```diff
@@ -0,0 +1,148 @@
...
+		s.client.
+			EXPECT().
+			SetServerBusy(3600). // endpoint defaults to 3600 seconds
+			Return(true, &model.Response{Error: nil}).
+			Times(1)
+
+		err := setBusyCmdF(s.client, &cobra.Command{}, []string{})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0], map[string]string{"status": "ok"})
```
larkox: The order of the Equals is wrong. The definition is Equal(expected, actual). Therefore, in this example, it should be:
`s.Require().Equal(map[string]string{"status": "ok"}, printer.GetLines()[0])`

This may look like a nitpick, but it shows on the logs whenever the test fails.

commands/system_test.go
```diff
@@ -0,0 +1,148 @@
...
+		const minutes = 15
+
+		s.client.
+			EXPECT().
+			SetServerBusy(minutes*60).
+			Return(true, &model.Response{Error: nil}).
+			Times(1)
+
+		err := setBusyCmdF(s.client, &cobra.Command{}, []string{strconv.Itoa(minutes * 60)})
+		s.Require().Nil(err)
```
larkox: Similar to a previous comment, we are preferring here NoError over Nil

commands/system.go
```diff
@@ -0,0 +1,96 @@
...
+	Example: `  system getbusy`,
+	RunE:    withClient(getBusyCmdF),
+}
+
+var SystemSetBusyCmd = &cobra.Command{
+	Use:     "setbusy [seconds]",
+	Short:   "Set the busy state to true",
+	Long:    `Set the busy state to true for the specified number of seconds, which disables non-critical services.`,
+	Example: `  system setbusy 3600`,
+	Args:    cobra.ExactArgs(1),
```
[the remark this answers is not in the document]
larkox: After seeing @mgdelacroix message, if we use flags this would not be necessary.

commands/system.go
wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/62adb429e8db58d66c95d6dd13a8bd538fe46bd7

commands/system.go
wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/62adb429e8db58d66c95d6dd13a8bd538fe46bd7

commands/system.go
wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/62adb429e8db58d66c95d6dd13a8bd538fe46bd7

commands/system.go
wiggin77: Yes, that's much cleaner.  I haven't used Cobra much - getting a crash course from you. :-)

Fixed via https://github.com/mattermost/mmctl/pull/130/commits/ea2d91a248e9056585d42f34458be0a3b48cb4de

commands/system.go
wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/ea2d91a248e9056585d42f34458be0a3b48cb4de

commands/system.go
wiggin77: Ugh, right.  I don't remember why I made that API return a bool and an error.  Seems redundant now.

wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/67d1566f460ead013fae90c82c7b37bb3e9f18a0    I double checked that the bool mirrors the error exactly (never returns false without error) so I'm just ignoring the bool.  Changing that API would be breaking so the behaviour won't change.  I don't know what I was thinking having that return a bool in the first place.

commands/system.go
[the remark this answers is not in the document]
wiggin77: Thanks.  Using flag now https://github.com/mattermost/mmctl/pull/130/commits/ea2d91a248e9056585d42f34458be0a3b48cb4de

commands/system_test.go
wiggin77: See above.  The bool is redundant and I'm now ignoring it.

commands/system_test.go
wiggin77: Good catch.  I'll fix that for the whole file.

commands/system_test.go
wiggin77: Seconds is now a flag, and parsing is handled by Cobra.  It returns error.

commands/system_test.go
wiggin77: I did implement this, per our sidebar conversation. I ended up discarding it because it was brittle was what we gain.  Some cases are simple, some not so much.  I did however fix up the cases where I was returning an error and printing an error.

commands/system_test.go
wiggin77: Fixed

commands/system_test.go
wiggin77: Fixed

commands/system_test.go
wiggin77: Fixed

commands/system.go
mgdelacroix: I think I've seen the pattern someplace else in our API. I agree with ignoring the error being a good solution

commands/system.go
```diff
@@ -0,0 +1,93 @@
...
+
+	sbs, response := c.GetServerBusy()
+	if response.Error != nil {
+		return fmt.Errorf("unable to get busy state: %w", response.Error)
+	}
+	printer.PrintT("busy:{{.Busy}} expires:{{.Expires}}", sbs)
+	return nil
+}
+
+func setBusyCmdF(c client.Client, cmd *cobra.Command, args []string) error {
```
mgdelacroix: We should add `printer.SetSingle(true)` to this command too

commands/system.go
```diff
@@ -0,0 +1,93 @@
...
+	_, response := c.SetServerBusy(int(seconds))
+	if response.Error != nil {
+		return fmt.Errorf("unable to set busy state: %w", response.Error)
+	}
+
+	printer.PrintT("Busy state set", map[string]string{"status": "ok"})
+	return nil
+}
+
+func clearBusyCmdF(c client.Client, cmd *cobra.Command, _ []string) error {
```
mgdelacroix: Same here

commands/system.go
wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/cb820ee21914e2fa5792285f7538644a9f79f523

commands/system.go
wiggin77: Fixed via https://github.com/mattermost/mmctl/pull/130/commits/cb820ee21914e2fa5792285f7538644a9f79f523

PR #131, opened 2020-03-05
Add coverage report

hanzei:


.codecov.yml
```diff
@@ -0,0 +1,6 @@
+comment: true
+coverage:
+  status:
+    project:
+      default:
+        informational: true
```
hanzei: This whole file might not be needed any longer if https://community-release.mattermost.com/core/pl/ay8eip538jnnmdzjfizxzrc77c get addressed, but I wanted to kick things off.

.codecov.yml
```diff
@@ -0,0 +1,6 @@
+comment: true
```
hanzei: There is no comment in this PR because there is no base to compare with

Makefile
```diff
@@ -85,6 +85,11 @@ test-all:
 	@echo Running all tests
 	$(GO) test -mod=vendor -race -v -tags 'unit e2e' $(GO_PACKAGES)
 
+.PHONY: coverage
+coverage:
+	$(GO) test -mod=vendor -race -tags unit -coverprofile=coverage.txt ./...
```
hanzei: I think it's fine to only use unit test coverage as it should be the same as the `e2e` one

hanzei:
You can see the report at https://codecov.io/gh/mattermost/mmctl/branch/mmctl

.circleci/config.yml
```diff
@@ -194,6 +211,7 @@ workflows:
     jobs:
       - docs
       - lint
+      - coverage
```
mgdelacroix: I think we should only run the `coverage` job if the tests pass:

```suggestion
      - coverage:
          requires:
            - test
```

Makefile
```diff
@@ -85,6 +85,11 @@ test-all:
 	@echo Running all tests
 	$(GO) test -mod=vendor -race -v -tags 'unit e2e' $(GO_PACKAGES)
 
+.PHONY: coverage
```
mgdelacroix: We can add this PHONY target to the top of the file

Makefile
mgdelacroix: I would add both tags to the coverage, as there are things that we are only testing in `e2e` (the client creation and connection for example)

.circleci/config.yml
hanzei: Is there any downside of running it anyway?

Makefile
hanzei: I like (2/5) having the `PHONY` next to the target as this makes it clear that every target as a `PHONY`. But I would change it, ff you feel strong about this.

Makefile
hanzei: The issue with also running the e2e test is that they (have to) run in docker. Which makes it a bit hard to extract the coverage report. Would you be fine just running the unit tests for now and I open a ticket for the e2e tests?

.circleci/config.yml
jespino: I think it can generate weird reports, imagine that something breaks during the tests (a panic) and only part of the tests has been run, probably the coverage is going to be not-realistic at all. I think I agree with @mgdelacroix here, I prefer only run coverage if tests passes.

Makefile
jespino: I think that is a good idea, opening another ticket to refine it later.

Makefile
jespino: I like consistency, my proposal is to put it at the top, and then, we can add another PR moving all .PHONY tags aside the tasks itself, but as part of another change.

hanzei: Opened https://github.com/mattermost/mmctl/issues/136

hanzei:
I forgot to add a badge. Will do later today.

PR #139, opened 2020-03-20
[MM-23482] Make team list iterate until it gets all server teams

mgdelacroix:
#### Summary
The `team list` command was querying for one page of teams with a `per-page` param of 10000 teams. The API has a limit of 200 for the `per-page` params, making the command only show the first 200.

This PR makes the command iterate over several teams pages, printing all the server teams.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23482


commands/team.go
```diff
@@ -176,17 +177,26 @@ func archiveTeamsCmdF(c client.Client, cmd *cobra.Command, args []string) error
...
-			printer.PrintT("{{.Name}}", team)
+		for _, team := range teams {
+			if team.DeleteAt > 0 {
+				printer.PrintT("{{.Name}} (archived)", team)
+			} else {
+				printer.PrintT("{{.Name}}", team)
+			}
 		}
+
+		if len(teams) < web.LIMIT_MAXIMUM {
```
wiggin77: Looks good. I double checked that teams endpoint doesn't error on empty page, otherwise if number of teams total was a multiple of `web.LIMIT_MAXIMUM` it would return error.  Instead, in that case, it will do one last request, get an empty list then exit clean.

wiggin77: Maybe add a unit test for team count a multiple of `web.LIMIT_MAXIMUM` just in case the endpoint (e.g. v6) starts erroring for page numbers beyond the last one?  Not sure if necessary.

commands/team.go
```diff
@@ -176,17 +177,26 @@ func archiveTeamsCmdF(c client.Client, cmd *cobra.Command, args []string) error
 }
 
 func listTeamsCmdF(c client.Client, cmd *cobra.Command, args []string) error {
-	teams, response := c.GetAllTeams("", 0, 10000)
-	if response.Error != nil {
-		return response.Error
-	}
+	page := 0
+	for {
+		teams, response := c.GetAllTeams("", page, web.LIMIT_MAXIMUM)
```
jespino: ok, importing this `web.LIMIT_MAXIMUM` constant is triggering here the inclusion in the vendor directory a zillion of new files. I think can be more interesting to have a copy of the constant and avoid all that vendors directory explosion.

commands/team.go
mgdelacroix: Hum... it's an interesting corner case, but I think that if the API would start erroring on empty pages, we would treat those errors as bugs, so the test case wouldn't be needed. I can add it if you feel strongly about it tho

PR #147, opened 2020-03-30
[GH-13461] add unit tests to the RemoveChannelUsersCmd mmctl command

fedealconada:
#### Summary
Adds unit tests to the RemoveChannelUsersCmd mmctl command

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/13461

### DIFF
--- a/commands/channel_test.go
+++ b/commands/channel_test.go
@@ -2181,15 +2181,14 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 	userArg := "user@example.com"
 	userID := "example-user-id"
 	mockUser := model.User{Id: userID, Email: userArg}
+	teamName := "teamName"
+	channelName := "channelName"
+	argsTeamChannel := teamName + ":" + channelName
 
 	s.Run("should remove user from channel", func() {
 		printer.Clean()
 
 		cmd := &cobra.Command{}
-
-		teamName := "teamName"
-		channelName := "channelName"
-		argsTeamChannel := teamName + ":" + channelName
 		args := []string{argsTeamChannel, "user@example.com"}
 
 		foundTeam := &model.Team{
@@ -2207,12 +2206,6 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 		s.client.
 			EXPECT().
 			GetTeam(teamName, "").
-			Return(nil, &model.Response{Error: nil}).
-			Times(1)
-
-		s.client.
-			EXPECT().
-			GetTeamByName(teamName, "").
 			Return(foundTeam, &model.Response{Error: nil}).
 			Times(1)
 
@@ -2236,7 +2229,6 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 
 		err := removeChannelUsersCmdF(s.client, cmd, args)
 		s.Require().Nil(err)
-		s.Require().Len(printer.GetErrorLines(), 0)
 		s.Require().Len(printer.GetLines(), 0)
 	})
 
@@ -2245,10 +2237,6 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 
 		cmd := &cobra.Command{}
 		cmd.Flags().Bool("all-users", true, "Remove all users from the indicated channel.")
-
-		teamName := "teamName"
-		channelName := "channelName"
-		argsTeamChannel := teamName + ":" + channelName
 		args := []string{argsTeamChannel, "user@example.com"}
 
 		err := removeChannelUsersCmdF(s.client, cmd, args)
@@ -2260,10 +2248,6 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 
 		cmd := &cobra.Command{}
 		cmd.Flags().Bool("all-users", true, "Remove all users from the indicated channel.")
-
-		teamName := "teamName"
-		channelName := "channelName"
-		argsTeamChannel := teamName + ":" + channelName
 		args := []string{argsTeamChannel}
 
 		foundTeam := &model.Team{
@@ -2284,12 +2268,6 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 		s.client.
 			EXPECT().
 			GetTeam(teamName, "").
-			Return(nil, &model.Response{Error: nil}).
-			Times(1)
-
-		s.client.
-			EXPECT().
-			GetTeamByName(teamName, "").
 			Return(foundTeam, &model.Response{Error: nil}).
 			Times(1)
 
@@ -2314,7 +2292,6 @@ func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
 		err := removeChannelUsersCmdF(s.client, cmd, args)
 
 		s.Require().Nil(err)
-		s.Require().Len(printer.GetErrorLines(), 0)
 		s.Require().Len(printer.GetLines(), 0)
 	})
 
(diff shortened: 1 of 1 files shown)


commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
 		s.Require().Equal(printer.GetLines()[0], updatedChannel)
 	})
 }
+
+func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
+	userArg := "user@example.com"
```
ethervoid: This should be userEmail to be more clear

commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+	mockUser := model.User{Id: userID, Email: userArg}
+	teamName := "teamName"
+	channelName := "channelName"
+	argsTeamChannel := teamName + ":" + channelName
+
+	s.Run("should remove user from channel", func() {
+		printer.Clean()
+
+		cmd := &cobra.Command{}
+		args := []string{argsTeamChannel, "user@example.com"}
```
ethervoid: You can `userEmail` instead of hardcode the email in the args

commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 0)
+	})
+
+	s.Run("should throw error if both --all-users flag and user email are passed", func() {
+		printer.Clean()
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Bool("all-users", true, "Remove all users from the indicated channel.")
+		args := []string{argsTeamChannel, "user@example.com"}
```
ethervoid: Same here

commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 0)
+	})
+
+	s.Run("should remove multiple users from channel", func() {
+		printer.Clean()
+
+		cmd := &cobra.Command{}
+		args := []string{argsTeamChannel, "user@example.com", "user@example.com"}
```
ethervoid: Same here

commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+
+		foundTeam := &model.Team{
+			Id:          "teamId",
```
ethervoid: Add this as constant and use it

commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
 		s.Require().Equal(printer.GetLines()[0], updatedChannel)
 	})
 }
+
+func (s *MmctlUnitTestSuite) TestRemoveChannelCmd() {
```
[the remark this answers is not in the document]
fedealconada: Yes, right, I agree with you and actually it was a bit confusing to me. But I named it like that because on `mattermost-server` repo ([here](https://github.com/mattermost/mattermost-server/blob/master/cmd/mattermost/commands/channel_test.go#L32)) I saw that the test function is called `TestRemoveChannel`... But I'll proceed with the suggestion and change it it that's okay.

commands/channel_test.go
fedealconada: Agree, again, I've named this variable like that because I've seen it already in other part of the code so to keep consistency, but I'll change it.

fedealconada:
Thanks for the feedback @ethervoid ! I've just pushed the changes

commands/channel_test.go
ethervoid: Please change all the `user@example.com` by the constant we have :)

commands/channel_test.go
```diff
@@ -2176,3 +2176,169 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+	argsTeamChannel := teamName + ":" + channelName
+
+	s.Run("should remove user from channel", func() {
+		printer.Clean()
+
+		cmd := &cobra.Command{}
+		args := []string{argsTeamChannel, "user@example.com"}
+
+		foundTeam := &model.Team{
+			Id:          "teamId",
```
ethervoid: These hardcoded values are duplicated all over the tests so we can move them to a constant and use those constants instead of the hardcoded values

fedealconada: I've added some constants and replace them over all the tests (not only the ones for RemoveChannelUsersCmd). If you prefer me not to touch the other tests, let me know and I can revert those changes and only use the constants for the RemoveChannelUsersCmd use cases. Let me know!

ethervoid: Great :), I prefer to have it that way. Let's wait for @mgdelacroix feedback in this case :)

mgdelacroix: Agree with the approach :)

commands/channel_test.go
```diff
@@ -1236,33 +1222,33 @@ func (s *MmctlUnitTestSuite) TestAddChannelUsersCmdF() {
 		s.Len(printer.GetLines(), 0)
 		s.Len(printer.GetErrorLines(), 0)
 	})
-	s.Run("Add existing user to existing channel", func() {
+	s.Run("Add existing user user@example.coto existing channel", func() {
```
mgdelacroix, line 1225: I think this change comes from a bad paste

```suggestion
	s.Run("Add existing user to existing channel", func() {
```

commands/channel_test.go
```diff
@@ -2176,3 +2142,165 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+		}
+
+		foundChannel := &model.Channel{
+			Id:          channelID,
+			Name:        channelName,
+			DisplayName: channelDisplayName,
+		}
+
+		mockMember := model.ChannelMember{ChannelId: channelID, UserId: mockUser.Id}
+		mockChannelMembers := model.ChannelMembers{mockMember, mockMember, mockMember}
```
mgdelacroix, line 2228: Can we change this members list to have three different members? I think the current one could eventually mask errors in the test

commands/channel_test.go
```diff
@@ -2176,3 +2142,165 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+
+		s.client.
+			EXPECT().
+			GetChannelMembers(foundChannel.Id, 0, 10000, "").
+			Return(&mockChannelMembers, &model.Response{Error: nil}).
+			Times(1)
+
+		s.client.
+			EXPECT().
+			RemoveUserFromChannel(foundChannel.Id, mockUser.Id).
```
mgdelacroix, line 2250: As per the above comment, this mock should be split in three, one per different user id

commands/channel_test.go
```diff
@@ -2176,3 +2142,165 @@ func (s *MmctlUnitTestSuite) TestRenameChannelCmd() {
...
+
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 0)
+	})
+
+	s.Run("should remove multiple users from channel", func() {
+		printer.Clean()
+
+		cmd := &cobra.Command{}
+		args := []string{argsTeamChannel, userEmail, userEmail}
```
mgdelacroix, line 2264: Same here, could we have two different users?

fedealconada:
done @mgdelacroix

PR #148, opened 2020-04-01
add unit tests to the UploadLicenseCmd mmctl command

fedealconada:
#### Summary
Adds unit tests to the UploadLicenseCmd mmctl command

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/13159


hanzei:
@fedealconada running `golangci-lint run --fix ./...` does resolve the import issue

fedealconada:
@hanzei ohh, I feel so dumb, I was trying to fix it with `gofmt -w ./commands/license_test.go` hehe, thanks!

commands/license.go
```diff
@@ -40,12 +40,14 @@ func init() {
 	RootCmd.AddCommand(LicenseCmd)
 }
 
+var customReadFile = ioutil.ReadFile // customReadFile can be reassigned to handle mocking
```
mgdelacroix, line 43: Although this is a clever approach to hijack the reading of the contents of the file, I think it's better to simply create a temporal file in the test and point the command to it, as it would be checking as well that it correctly reads the filename passed to it, and we don't need to modify the business logic of the command itself.

You can use the way that the [plugin add command test mocks the plugin file](https://github.com/mattermost/mmctl/blob/master/commands/plugin_test.go#L23) as a reference.

fedealconada: @mgdelacroix I kind of got inspired by [this](https://endler.dev/2018/go-io-testing/). If you want me to anyways do the test by creating a temporary file, I'll do it, no problem at all. Let me know your thoughts! :)

mgdelacroix: @fedealconada Interesting! Thanks for pointing to the article :)

After reading the article, I understand your rationale here, and I still think that there are reasons to go for the temp file:

Luckily for us, we can use temp files (hence we don't have to deal with the constrained environment problem) and it's very unlikely that the CI runs into I/O problems (if it does, reading files would not be our main issue I'm afraid :P), those being the main two problems that the blogpost approach is trying to address.

On the other hand, the temp file approach is simple, doesn't need for us to modify the function's business logic (which not mandatory but desirable), helps testing as well the path argument usage and is consistent with the current codebase.

fedealconada: Great! Makes sense @mgdelacroix :) I'll do that change then! Thankss

fedealconada: Done @mgdelacroix

hanzei:
@mgdelacroix @ethervoid Is this PR ready to merge?

PR #157, opened 2020-04-20
MM-24294: Check the type before assigning a value to a slice

agnivade:
#### Summary

﻿(reflect.Value).Set was being called without checking whether the input
value's type was indeed a slice or not. This would cause a panic in case
when the user is unsure of how to enter an array value to set a config key.

To prevent this, we check for the type and raise an appropriate error
and also add an example to show how to set an array value.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24294

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -42,7 +42,7 @@ var ConfigSetCmd = &cobra.Command{
 	Use:     "set",
 	Short:   "Set config setting",
 	Long:    "Sets the value of a config setting by its name in dot notation. Accepts multiple values for array settings",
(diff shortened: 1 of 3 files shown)


lindy65:
Hi @agnivade - There is no option to set up a spinwick in the `mmctl` repo. Should this option be added so I can test against this PR? Or is it fine to test this post-merge?

agnivade:
Hey @lindy65 - it's fine to test post-merge. There's already a unit test which verifies this behavior.

PR #158, opened 2020-05-11
Migrates the make_private command to use the update channel privacy endpoint

mgdelacroix:
#### Summary
To allow for deprecation of the [convert channel to private endpoint](https://api.mattermost.com/#tag/channels/paths/~1channels~1{channel_id}~1convert/post) this PR migrates its usage to the [update channel privacy](https://api.mattermost.com/#tag/channels/paths/~1channels~1{channel_id}~1privacy/put) one

Jira ticket: https://mattermost.atlassian.net/browse/MM-25048

### DIFF
--- a/commands/channel.go
+++ b/commands/channel.go
@@ -358,7 +358,7 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string) 
 		return errors.New("you can only change the type of public channels")
 	}
 
-	if _, response := c.ConvertChannelToPrivate(channel.Id); response.Error != nil {
+	if _, response := c.UpdateChannelPrivacy(channel.Id, model.CHANNEL_PRIVATE); response.Error != nil {
 		return response.Error
 	}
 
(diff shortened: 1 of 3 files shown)


commands/channel.go
```diff
@@ -358,7 +358,7 @@ func makeChannelPrivateCmdF(c client.Client, cmd *cobra.Command, args []string)
 		return errors.New("you can only change the type of public channels")
 	}
 
-	if _, response := c.ConvertChannelToPrivate(channel.Id); response.Error != nil {
```
ethervoid, line 361: Why are we making this change?

mgdelacroix: We used to have an endpoint to make a channel private (and the `ConvertChannelToPrivate` was making use of it) and then we implemented the `UpdateChannelPrivacy` one, which allows to change privacy in both ways. First one is not deprecated yet, but we rather use the new one to allow for deprecation

ethervoid: Ok!

PR #162, opened 2020-05-20
Disable syscall usage for windows build

mgdelacroix:
#### Summary
This PR creates two versions of the `checkValidSocket` func, as windows builds were failing due to `syscall.Stat_t` not being available for the OS.

### DIFF
--- a/commands/utils_unix.go
+++ b/commands/utils_unix.go
@@ -1,0 +1,41 @@
+// +build linux darwin
+
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package commands
+
+import (
+	"fmt"
+	"os"
+	"os/user"
+	"syscall"
+)
+
+func checkValidSocket(socketPath string) error {
+	// check file mode and permissions
+	fi, err := os.Stat(socketPath)
+	if err != nil && os.IsNotExist(err) {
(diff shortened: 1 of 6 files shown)


commands/utils_unix.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// +build linux darwin
```
agnivade, line 4: Is this still needed if the file ends with "_unix.go" ? As per: https://golang.org/pkg/go/build/#hdr-Build_Constraints

mgdelacroix: I'm afraid that it is. Only vars allowed in `GOOS` are allowed as suffix, so `_linux.go` or `_darwin.go` would work, but AFAIK there is no value that would match both at the same time

agnivade: I mean that we just want to exclude windows right ? So it's essentially windows and !windows. So isn't the _unix.go and _windows.go file extensions sufficient to make this work ?

mgdelacroix: I'm not sure I'm understanding your proposal. If we remove the build tags from the `_unix.go` file, it would get included in the windows build (as the `_unix` suffix has no effect because is not a valid `GOOS` value) and that would make the windows build fail

agnivade: Never mind. I got it now. Thanks.

PR #173, opened 2020-07-08
[MM-23582] list private channels in channel list

ashishbhate:
#### Summary

- List the private channels that the user is a member of or has the permissions to list in the channel list command
- Updates the mattermost-server version in the module to point to the version with support for this feature.

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-23582


commands/channel.go
```diff
@@ -345,6 +348,14 @@ func listChannelsCmdF(c client.Client, cmd *cobra.Command, args []string) error
 		for _, channel := range deletedChannels {
 			printer.PrintT("{{.Name}} (archived)", channel)
 		}
+
+		privateChannels, err := getPrivateChannels(c, team.Id)
+		if err != nil {
+			printer.PrintError("Unable to list private channels for '" + args[i] + "'. Error: " + err.Error())
```
wiggin77, line 354: args[i] may not match teams[i] since any teams provided on the command line that do not exist will be simply skipped.  I suggest using `team.Name` here.

ashishbhate: Teams provided on the command line that do not exist are returned as "nil" by `getTeamsFromTeamArgs`. So `args[i]` will always match `teams[i]`

The check if a `team` is `nil` is done on line 330, so there is no risk of nil pointer deference either.

I've used `args[i]` here to be consistent with the error messages in other parts of the code. I think the usage of `args[i]` gives clearer feedback to the user that the entered argument is bad. It may not be clear from the team name which corresponding input team id was bad, especially in a long argument list.

PR #178, opened 2020-07-20
Use client4 socket creation function

mgdelacroix:


mgdelacroix:
CI is expected to fail until https://github.com/mattermost/mattermost-server/pull/15067 is merged and we change the commit that this PR is using so the function becomes available

lieut-data:
The underlying change here is one of the coolest things I've seen in a while -- nice :D

PR #181, opened 2020-07-20
MM-25543: Adds commands to assign/unassign users to system roles.

mkraft:
#### Summary

Adds commands to assign and unassign system roles to users by username.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25543


mkraft:
> Looks good to me, but is there somewhere that one can go to get the full list of roles without looking at the server code?

Not currently, but I think that's a good feature to add, particularly with the release of custom roles. I think we're OK to document it for now because the feature is in beta and there are only four roles with which to use `permissions assign` and `permissions unassign`. Because of your comment I have just added all current examples to the docs.

mkraft:
@mgdelacroix Perhaps when you review this you could give me your opinion on whether these new commands should remain under `permissions` or be a subcommand like `permissions roles`.

commands/permissions.go
```diff
@@ -44,11 +46,37 @@ var ShowRoleCmd = &cobra.Command{
 	RunE:    withClient(showRoleCmdF),
 }
 
+var AssignUsersCmd = &cobra.Command{
+	Use:   "assign [role_name] [username...]",
```
mgdelacroix, line 50: I think that, as you commented in the PR thread, this belongs to its own subcommand `permission role assign`. Right now it could sound like assigning permissions to users, and as these commands operate over the role entity, I think it makes total sense to create a `role` subcommand

commands/permissions.go
```diff
@@ -44,11 +46,37 @@ var ShowRoleCmd = &cobra.Command{
...
+	Short: "Assign users to role (EE Only)",
+	Long:  "Assign users to a role by username (Only works in Enterprise Edition).",
+	Example: `  permissions assign system_admin john.doe jane.doe
```
mgdelacroix, line 53: nit: could you please add some comments to the examples? something like what we have in the [plugin marketplace ones](https://github.com/mattermost/mmctl/blob/master/commands/plugin_marketplace.go#L40)

commands/permissions.go
```diff
@@ -44,11 +46,37 @@ var ShowRoleCmd = &cobra.Command{
...
+  permissions assign system_read_only_admin john.doe jane.doe`,
+	Args: cobra.MinimumNArgs(2),
+	RunE: withClient(assignUsersCmdF),
+}
+
+var UnassignUsersCmd = &cobra.Command{
+	Use:   "unassign [role_name] [username...]",
+	Short: "Unassign users from role (EE Only)",
+	Long:  "Unassign users from a role by username (Only works in Enterprise Edition).",
+	Example: `  permissions unassign system_admin john.doe jane.doe
```
mgdelacroix, line 65: Same comments in the examples suggestion

commands/permissions.go
```diff
@@ -132,3 +160,60 @@ Scheme Managed: no
...
 }
+
+func assignUsersCmdF(c client.Client, cmd *cobra.Command, args []string) error {
+	role, response := c.GetRoleByName(args[0])
+	if response.Error != nil {
+		return response.Error
+	}
+
+	for _, username := range args[1:] {
+		user, response := c.GetUserByUsername(username, "")
```
mgdelacroix, line 171: here we can use the `getUsersFromUserArgs` to allow for more ways to reference users

commands/permissions.go
```diff
@@ -132,3 +160,60 @@ Scheme Managed: no
...
+			return response.Error
+		}
+	}
+
+	return nil
+}
+
+func unassignUsersCmdF(c client.Client, cmd *cobra.Command, args []string) error {
+	for _, username := range args[1:] {
+		user, response := c.GetUserByUsername(username, "")
```
mgdelacroix, line 195: Same `getUsersFromUserArgs` suggestion

commands/permissions.go
mkraft: Then we should move the `permissions show [role_name]` over to `permissions role show [role_name]` too.

mkraft: Agreed and updated. Thanks.

commands/permissions.go
mkraft: Added some comments.

commands/permissions.go
mkraft: Added.

commands/permissions.go
mkraft: Switched to it. Thanks.

commands/permissions.go
mkraft: Updated.

commands/permissions.go
```diff
@@ -132,3 +177,54 @@ Scheme Managed: no
...
+
+func assignUsersCmdF(c client.Client, cmd *cobra.Command, args []string) error {
+	role, response := c.GetRoleByName(args[0])
+	if response.Error != nil {
+		return response.Error
+	}
+
+	users := getUsersFromUserArgs(c, args[1:])
+
+	for _, user := range users {
```
mgdelacroix, line 189: we should check that all users have been found when iterating, maybe adding something like:

```suggestion
	for i, user := range users {
		if user == nil {
			printer.PrintError("Couldn't find user '" + args[i+1] + "'")
			continue
		}
```

commands/permissions.go
```diff
@@ -132,3 +177,54 @@ Scheme Managed: no
...
+		}
+	}
+
+	return nil
+}
+
+func unassignUsersCmdF(c client.Client, cmd *cobra.Command, args []string) error {
+	users := getUsersFromUserArgs(c, args[1:])
+
+	for _, user := range users {
```
mgdelacroix, line 210: Same suggestion here:

```suggestion
	for i, user := range users {
		if user == nil {
			printer.PrintError("Couldn't find user '" + args[i+1] + "'")
			continue
		}
```

mgdelacroix (changes requested):

commands/permissions.go
mkraft: I've update it, thanks.

commands/permissions.go
```diff
@@ -36,12 +35,13 @@ var RemovePermissionsCmd = &cobra.Command{
 }
 
 var ShowRoleCmd = &cobra.Command{
```
mgdelacroix, line 37: this command has been moved, so we can delete it from this file

commands/permissions_role.go
```diff
@@ -0,0 +1,165 @@
...
+	for i, user := range users {
+		if user == nil {
+			printer.PrintError("Couldn't find user '" + args[i+1] + "'.")
+			continue
+		}
+
+		userRoles := strings.Fields(user.Roles)
+		originalCount := len(userRoles)
+
+		for i := 0; i < len(userRoles); i++ {
```
mgdelacroix, line 149: we're reusing `i` here from the outer loop, so this loop needs to be changed to use `j` or a similar name

commands/permissions.go
mkraft: I thought we wanted this to continue working using the old `permissions show [role_name]` command, but with a deprecation warning.

commands/permissions_role.go
mkraft: Good point.

commands/permissions.go
mgdelacroix: You're completely right, I oversaw the deprecation property, sorry :facepalm:

mgdelacroix (approved):

PR #184, opened 2020-07-31
Bump golang build image

metanerd:


agnivade:
Miguel is on leave. And this is fine not to be backported, since mmctl is not affected by the bug. Leaving only for master.

PR #194, opened 2020-08-26
[MM-28098] Add user change-password command

mgdelacroix:
#### Summary
This PR adds the `user change-password` command that allows a user to change their own password or an admin to change any user's password.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28098


lindy65 (approved):
Thank @mgdelacroix - this will be tested after merging 👍

commands/user.go
```diff
@@ -67,14 +67,36 @@ var SendPasswordResetEmailCmd = &cobra.Command{
...
 	Short:   "Change email of the user",
 	Long:    "Change email of the user.",
 	Example: "  user email testuser user@example.com",
 	RunE:    withClient(updateUserEmailCmdF),
 }
 
+var ChangePasswordUserCmd = &cobra.Command{
+	Use:   "change-password <user>",
+	Short: "Changes a user's password",
+	Long:  "Changes the password of a user by a new one provided. If the user is changing it's own password, the flag --current must indicate the current password. The flag --hashed can be used to indicate that the new password has been introduced already hashed",
```
gabrieljackson, line 81: nit:
```suggestion
	Long:  "Changes the password of a user by a new one provided. If the user is changing their own password, the flag --current must indicate the current password. The flag --hashed can be used to indicate that the new password has been introduced already hashed",
```

commands/user.go
```diff
@@ -372,6 +399,51 @@ func updateUserEmailCmdF(c client.Client, cmd *cobra.Command, args []string) err
...
+			if err != nil {
+				return errors.New("couldn't read password: " + err.Error())
+			}
+		}
+
+		fmt.Printf("New password: ")
+		var err error
+		password, err = getPasswordFromStdin()
+		if err != nil {
+			return errors.New("couldn't read password: " + err.Error())
```
gabrieljackson, line 424: Just curious if there are plans to use `errors.Wrap()` at any point from `github.com/pkg/errors`? I see that it is imported, so was just wondering why we didn't use it.

mgdelacroix: Yep, it makes total sense to wrap the errors in this case. Cobra doesn't do anything special with the wrapped information, but there is no reason not to do it really

Updated where possible :)

commands/user.go
```diff
@@ -372,6 +399,51 @@ func updateUserEmailCmdF(c client.Client, cmd *cobra.Command, args []string) err
...
+	}
+
+	user := getUserFromUserArg(c, args[0])
+	if user == nil {
+		return errors.New("couldn't find user '" + args[0] + "'")
+	}
+
+	if hashed {
+		if _, resp := c.UpdateUserHashedPassword(user.Id, password); resp.Error != nil {
+			return errors.Wrap(resp.Error, "changing user password failed")
```
ethervoid, line 435: I'd change this to `changing hashed user password failed` so we can differentiate it from the other error

commands/user.go
```diff
@@ -67,14 +67,36 @@ var SendPasswordResetEmailCmd = &cobra.Command{
...
+  # if you are changing your own password, you need to provide the current one
+  $ mmctl user change-password my-username --current current-password --password new-password
+
+  # you can ommit these flags to introduce them interactively
+  $ mmctl user change-password my-username
+  Are you changing your own password? (YES/NO): YES
+  Current password:
+  New password:
+
+  # if you have system permissions, you can update the password with the already hashed new password
```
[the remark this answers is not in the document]
mgdelacroix: Updated! PTAL

ethervoid (approved):
LGTM :bow:

PR #198, opened 2020-09-07
Update go and golangci versions

mgdelacroix:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/mattermost/mmctl
 
-go 1.13
+go 1.14
```
agnivade, line 3: Wondering why is this needed? Unless we are raising the minimum required Go version to compile mmctl (which I don't think we are), there's no need to change this.

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ executors:
   default:
     working_directory: ~/go/src/github.com/mattermost/mmctl
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/golang:1.14.6
```
agnivade, line 10: Let's go for 1.14.8

agnivade (commented):
Just some small comments.

go.mod
mgdelacroix: Right! Changed back to `1.13`

isacikgoz (approved):
LGTM 👍

agnivade (approved):

metanerd (approved):
:+1:

PR #220, opened 2020-10-02
[MM-28433] introduce new xdg config location and ability to specify c…

muety:
…onfig path via flag or env variable

#### Summary
* Always save new config files to `~/.config/mmctl`
* Always try to read config file in the following precedence order:
  1. The path specified through ~~`-c`,~~ `--config` or `$MMCTL_CONFIG`
  2. The above default path
  3. The "legacy" path at `~/` as a fallback

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15679

#### Please Note
I thought of including tests to verify the above order is kept and config files are read and written properly. But this would require to mock file system calls performed by `os` packages and I am not aware of a good way to do so.


hanzei:
Thanks for the PR @muety :+1: I looks lint CI broke because of the changes. Could you please take a look at  https://app.circleci.com/pipelines/github/mattermost/mmctl/1547/workflows/2bbdeee0-9554-4845-89ef-e748a1cf1c12/jobs/5320/parallel-runs/0/steps/0-102?

muety:
Apparently, docs generation is failing, but I can't figure out the reason. Any hints?

docs/mmctl.rst
```diff
@@ -16,6 +16,7 @@ Options
 
 ::
 
+      --config string                path to search for '.mmctl' configuration file (default "/home/ferdinand/.config/mmctl")
```
hanzei, line 19: The default path depends on who is running the command. Can you change this to e.g. `$HOME/.config/mmctl`?

muety: Sorry, my bad, I should have seen this!

commands/auth_utils.go
```diff
@@ -29,25 +37,41 @@ type Credentials struct {
...
+
+	configPath := strings.Replace(viper.GetString("config"), "$HOME", u.HomeDir, 1)
+
+	// probe user-defined or xdg-compliant default config path
+	if _, err := os.Stat(filepath.Join(configPath, configFileName)); err != nil {
+		if !os.IsNotExist(err) {
+			return "", err
+		}
+		// conditionally fall back to legacy config path
+		configPath = u.HomeDir
```
ethervoid, line 58: Probably we want to show a warning or message if that happens giving feedback to the user

ethervoid (approved):
Great job! Just a minor comment. Regarding the tests, that would be awesome. You can just define the flag and clean up the files after that but probably Miguel could help you better

commands/auth_utils.go
muety: Sure!

muety: > Regarding the tests, that would be awesome. You can just define the flag and clean up the files after that but probably Miguel could help you better

@mgdelacroix Do you have a good solution for how to test this? I'd like to avoid creating and deleting real files in the user's actual home directory. It would be fine for CI/CD, but if tests are run in a developers' machine, her local config might potentially be overwritten / deleted in the worst case.

mgdelacroix: I think that the best option here is to use something like:

```go
var currentUser *user.User

func SetUser(newUser *user.User) {
	currentUser := newUser
}

func init() {
	newUser, err := user.Current()
	if err != nil {
		panic(err)
	}

	SetUser(newUser)
}
```

This way you can use `SetUser` at the beginning of the tests to mock a fake `HomeDir` like this:

```go
dir, _ := ioutil.TempDir("", "mmctl-")
defer os.RemoveAll(dir)

u, err := user.Current()
if err != nil {
	panic(err)
}

oldUser = currentUser
u.HomeDir = dir
SetUser(u)
defer func () {
    SetUser(oldUser)
}
```

mgdelacroix (changes requested):
Very nice work @muety! Requested some changes mostly for tests and to take some cases into account, but it's looking really good!!!

commands/auth_utils.go
```diff
@@ -29,25 +37,41 @@ type Credentials struct {
 
 type CredentialsList map[string]*Credentials
 
-func getConfigFilePath() (string, error) {
-	user, err := user.Current()
+func getDefaultConfigPath() string {
+	return filepath.Join("$HOME", xdgConfigHome, "mmctl")
```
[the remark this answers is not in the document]
muety: The reason I did this was to provide a proper help text in `root.go`, where the flags are defined. Otherwise, all docs generated on my machine, for instance, would contain the default value of `/home/ferdinan/.config` (i.e. my local username) for this argument.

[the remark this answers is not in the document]
muety: The reason I did this was to provide a proper help text in `root.go`, where the flags are defined. Otherwise, all docs generated on my machine, for instance, would contain the default value of `/home/ferdinan/.config` (i.e. my local username) for this argument.

muety:
@mgdelacroix Thanks for your help! I implemented your requested changes and also added basic tests. Feel free to request more changes if you're not 100 % comfortable with the code, I'm happy to learn about best practices.

hanzei:
@muety Heads up that the build is currently failing. Could you please take a look?

muety:
Apparently, `testing.T.TempDir` was only [introduced in Go `1.15`](https://golang.org/doc/go1.15#testing), but the pipeline runs on `1.14.8`. Everything should be fine now.

muety:
What else is left to do here?

mgdelacroix (approved):
Apologies for the delay on the review, LGTM! Thanks a lot for your contribution @muety!

ethervoid (approved):
LGTM!

lindy65 (approved):
Thanks @muety @mgdelacroix - looks good to me 👍

ethervoid:
Thank you for the hard work @muety :)

muety:
Reviewing this was probably harder 😃 Thanks for guiding my through my first Mattermost PR.

PR #221, opened 2020-10-03
[GH-15734] add e2e test for 'listChannelsCmdF' (#15734)

Oppodelldog:
#### Summary
This adds a e2e test for 'listChannelsCmdF'

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/15734


commands/user_e2e_test.go
```diff
@@ -136,3 +136,41 @@ func (s *MmctlE2ETestSuite) TestSearchUserCmd() {
...
+func (s *MmctlE2ETestSuite) TestListChannelsCmdF() {
+	s.SetupTestHelper().InitBasic()
+
+	s.RunForAllClients("List channels", func(c client.Client) {
+		printer.Clean()
+
+		err := listChannelsCmdF(c, &cobra.Command{}, []string{s.th.BasicTeam.Name})
+		s.Require().Nil(err)
+
+		validChannelNames := append(
```
Oppodelldog, line 149: I am not sure about the detail level of this test.

Depending on the client, the number of returned channels is different:
  6 for the normal client
  7 for local and sysadmin

To not test the details as requested in the ticket description I went for building this list of valid names.

The assertions then check that "all lines returned are channels and their names are in the valid set".

agnivade: Hey @Oppodelldog - I think this can be easily done if you use the `s.RunForSystemAdminAndLocal` for one case, and use `s.Run` for the other.

agnivade (commented):
Looks great @Oppodelldog. Just a small suggestion to improve the test further.

agnivade:
Sorry, I forgot, we need to add this test in `commands/channel_e2e_test.go`, not `commands/user_e2e_test.go`.

Oppodelldog:
> Sorry, I forgot, we need to add this test in `commands/channel_e2e_test.go`, not `commands/user_e2e_test.go`.

sure, thanks for the guidance, 7316514
I also added the license then with 4cf8370

commands/channel_e2e_test.go
```diff
@@ -0,0 +1,63 @@
...
+		team := "non-existent-team"
+
+		err := listChannelsCmdF(c, &cobra.Command{}, []string{team})
+		s.Require().Nil(err)
+		s.Len(printer.GetErrorLines(), 1)
+		s.Equal("Unable to find team '"+team+"'", printer.GetErrorLines()[0])
+	})
+}
+
+func (s *MmctlE2ETestSuite) assertOutputDisplaysChannels() {
```
mgdelacroix, line 48: I like the approach of building a list, but I'd avoid to make it a function on the test suite. Would you be open to create the list at the beginning of the `TestListChannelsCmdF` function and run one loop per test case??

For completion, in the test case of the unprivileged client with 6 channels in the response, I'd assert as well that the 7nth channel is not part of the list.

mgdelacroix (changes requested):
Thanks for your contribution @Oppodelldog, looking really good!! :rocket: 

Added only one change request

agnivade (approved):
Nothing on top of @mgdelacroix's suggestions. LGTM.

lindy65 (approved):
Thanks @Oppodelldog - looks good to me 👍

commands/channel_e2e_test.go
mgdelacroix: Thanks a lot for this change, I think a function inside the scope of the tests makes a lot of sense.

mgdelacroix (approved):
LGTM! Thanks a lot @Oppodelldog!! :+1:

PR #223, opened 2020-10-04
Add e2e tests for archiveTeamsCmdF

shihanng:
Close https://github.com/mattermost/mattermost-server/issues/15671


commands/team_e2e_test.go
```diff
@@ -51,3 +53,41 @@ func (s *MmctlE2ETestSuite) TestRenameTeamCmdF() {
...
+	s.RunForSystemAdminAndLocal("Archive basic team", func(c client.Client) {
+		printer.Clean()
+
+		err := archiveTeamsCmdF(c, cmd, []string{s.th.BasicTeam.Name})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		team := printer.GetLines()[0].(*model.Team)
+		s.Require().Equal(s.th.BasicTeam.Name, team.Name)
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
```
mgdelacroix, line 82: We should retrieve the team from the database using `s.th.App` and assert that after the command's execution it is now archived. After that, the team should be unarchived so the rest of the tests start from the right state.

I'm thinking on something like:

```go
basicTeam, err := s.th.App.GetTeam(s.th.BasicTeam.Id)
s.Require().Nil(err)
s.Require().NonZero(basicTeam.DeleteAt)

err := s.th.App.RestoreTeam(s.th.BasicTeam.Id)
s.Require().Nil(err)
```

commands/team_e2e_test.go
```diff
@@ -51,3 +53,41 @@ func (s *MmctlE2ETestSuite) TestRenameTeamCmdF() {
...
+
+	s.Run("Archive team without permissions", func() {
+		printer.Clean()
+
+		err := archiveTeamsCmdF(s.th.Client, cmd, []string{s.th.BasicTeam.Name})
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 1)
+		s.Require().Contains(printer.GetErrorLines()[0], "You do not have the appropriate permissions.")
+	})
```
mgdelacroix, line 92: Same here, could you please add a check to fetch the team from the database and assert that it is not archived?

mgdelacroix (changes requested):
Looking really good @shihanng!! Added a couple of comments to add checks related to the final database status

shihanng:
@mgdelacroix , thank you for the reviews. I've addressed the comments.

mgdelacroix (approved):
LGTM! :rocket: thanks for your contribution @shihanng!!

ethervoid (approved):
LGTM! Thank you for your contribution!

lindy65 (approved):
Thanks @shihanng @mgdelacroix - looks good to me 👍

PR #226, opened 2020-10-05
[MM-29230] Add e2e tests to teamUsersAddCmdF

simross:
**Summary**

Adds e2e tests for `teamUsersAddCmdF`.

**Ticket Link**

Fixes mattermost/mattermost-server#15731

JIRA: https://mattermost.atlassian.net/browse/MM-29230


commands/team_users_e2e_test.go
```diff
@@ -0,0 +1,139 @@
+package commands
```
jespino, line 1: You have to add the license header here :) (Just copy it from any other file in the project)

commands/team_users_e2e_test.go
```diff
@@ -0,0 +1,139 @@
...
+		s.Require().NotNil(teamMembers)
+		s.Require().Len(teamMembers, 2)
+
+		var teamUsersID []string
+		for _, v := range teamMembers {
+			teamUsersID = append(teamUsersID, v.UserId)
+		}
+		s.Require().Contains(teamUsersID, user.Id)
+
+		appErr = unlinkUserFromTeam(team.Id, s.th.BasicUser.Id)
```
jespino, line 110: Should we defer this right after adding the member? Failing this tests can end up cascading in other failing tests because the database has this user.

jespino (changes requested):
A couple of tiny things, but otherwise, looks good to me. Thanks @simross! :tada:

jespino (approved):
Looks good to me. Thanks @simross! :tada:

jespino:
@jasonblais I don't think QA is needed for this PR. This is only affecting new tests to be added. There is nothing to QA.

jasonblais:
@jespino okay, removing QA review label

commands/team_users_e2e_test.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+package commands
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
```
agnivade, line 8: Typically we place 3rd party imports after our own imports.

commands/team_users_e2e_test.go
```diff
@@ -0,0 +1,142 @@
...
+		s.Require().Len(teamMembers, 2)
+
+		var teamUsersID []string
+		for _, v := range teamMembers {
+			teamUsersID = append(teamUsersID, v.UserId)
+		}
+		s.Require().Contains(teamUsersID, user.Id)
+	})
+
+	s.RunForSystemAdminAndLocal("Add user to nonexistent team", func(c client.Client) {
```
agnivade, line 115: I think these 2 tests make sense for `s.RunForAllClients`?

commands/team_users_e2e_test.go
```diff
@@ -0,0 +1,142 @@
...
+		DisplayName: "dn_" + model.NewId(),
+		Name:        api4.GenerateTestTeamName(),
+		Email:       s.th.GenerateTestEmail(),
+		Type:        model.TEAM_OPEN,
+	})
+	s.Require().Nil(appErr)
+
+	unlinkUserFromTeam := func(teamId string, userId string) error {
+		teamMembers, err := s.th.App.GetTeamMembers(teamId, 0, 10, nil)
+
```
agnivade, line 33: Nit: the empty line can be removed.

agnivade (commented):
Thanks for the PR @simross ! I just have some minor suggestions.

commands/team_users_e2e_test.go
simross: Currently there are 2 cases, which have similar output:

* given team not found
* user doesn't have permission to view given team

In second case debug message is being printed, but there is no obvious way to reasonably check this during tests.  Therefore first test can be misleading if it is being run by usual user.

simross: Usual user need to be added to team in order for second test to work correctly. If combined in one test - also demands detection of user type during execution, which seems clunky. For readability separate test was created to cover tested case.

agnivade: Makes sense. Thanks.

agnivade (approved):
Thanks!

PR #231, opened 2020-10-07
[MM-29166] Add e2e tests to listTeamsCmdF

sirMackk:
Add end to end tests to listTeamsCmdF.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15672

JIRA: https://mattermost.atlassian.net/browse/MM-29166

### DIFF
--- a/commands/team_e2e_test.go
+++ b/commands/team_e2e_test.go
@@ -4,9 +4,9 @@
 package commands
 
 import (
-	"github.com/spf13/cobra"
-
 	"github.com/mattermost/mattermost-server/v5/model"
+
+	"github.com/spf13/cobra"
 
 	"github.com/mattermost/mmctl/client"
 	"github.com/mattermost/mmctl/printer"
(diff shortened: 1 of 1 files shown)


reflog (approved):

commands/team_e2e_test.go
```diff
@@ -51,3 +53,40 @@ func (s *MmctlE2ETestSuite) TestRenameTeamCmdF() {
...
+		err := listTeamsCmdF(c, &cobra.Command{}, []string{})
+		s.Require().Nil(err)
+		s.Len(printer.GetLines(), 2)
+		team := printer.GetLines()[0].(*model.Team)
+		s.Equal(s.th.BasicTeam.Name, team.Name)
+	})
+
+	s.RunForSystemAdminAndLocal("Should print archived teams", func(c client.Client) {
+		printer.Clean()
+		printer.SetFormat(printer.FormatPlain)
```
mgdelacroix, line 76: Would you be open to use the default format instead of plain and do the same type assertion style checks like you're doing in the first test case? It is much more robust to check on the returned structs than on the plaintext messages, so we try to do it if we can

commands/team_e2e_test.go
```diff
@@ -51,3 +53,40 @@ func (s *MmctlE2ETestSuite) TestRenameTeamCmdF() {
...
+		defer printer.SetFormat(printer.FormatJSON)
+
+		err := listTeamsCmdF(c, &cobra.Command{}, []string{})
+		s.Require().Nil(err)
+		s.Len(printer.GetLines(), 2)
+		s.Equal(printer.GetLines()[1], mockTeamName+" (archived)")
```
mgdelacroix, line 82: Actually, both calls use the same command, so I'd suggest to assert both as part of the first test case and change its name to something like "Should print both active and archived teams for syasdmin and local clients"

mgdelacroix (changes requested):
Thanks for the PR @sirMackk!! Looking really good, a suggestion around assertion organisation added

commands/team_e2e_test.go
sirMackk: Thanks for the feedback - I compressed both cases into one using the model struct instead of checking the text output.

lindy65 (approved):
Thanks @sirMackk - looks good to me 👍

mgdelacroix (approved):
Thanks for the changes @sirMackk, LGTM! :+1:

PR #250, opened 2020-10-18
MM-29563 added e2e tests to "modifyCommandCmdF"

JeremyShih:
#### Summary

Add end to end tests to the "modifyCommandCmdF"
#### Ticket Link

Fixes mattermost/mattermost-server#15941
[jira](https://mattermost.atlassian.net/browse/MM-29563)

### DIFF
--- a/commands/command_e2e_test.go
+++ b/commands/command_e2e_test.go
@@ -263,15 +263,23 @@ func (s *MmctlE2ETestSuite) TestModifyCommandCmdF() {
 	}
 
 	command, _ := s.th.SystemAdminClient.CreateCommand(newCmd)
+	copy := command
+	originalURL := newCmd.URL
 
 	s.RunForAllClients("modifyCommandCmdF", func(c client.Client) {
 		printer.Clean()
+
+		command = copy
+		s.Require().Equal(command.URL, originalURL)
+
 		// Reset the cmd and parse to force Flag.Changed to be true.
 		cmd := CommandModifyCmd
 		cmd.ResetFlags()
 		addCommandFieldsFlags(cmd)
+		url := command.URL + "_modified"
 		err := cmd.ParseFlags([]string{
 			command.Id,
+			"--url=" + url,
 		})
 		s.Require().Nil(err)
 
@@ -279,6 +287,9 @@ func (s *MmctlE2ETestSuite) TestModifyCommandCmdF() {
 		s.Require().Nil(err)
 		s.Len(printer.GetLines(), 1)
 		s.Len(printer.GetErrorLines(), 0)
+
+		check := getCommandFromCommandArg(s.th.SystemAdminClient, command.Id)
+		s.Require().Equal(check.URL, url)
 	})
 
 	s.RunForAllClients("modifyCommandCmdF for command that does not exist", func(c client.Client) {
--- a/commands/command_e2e_test.go
+++ b/commands/command_e2e_test.go
@@ -289,7 +289,9 @@ func (s *MmctlE2ETestSuite) TestModifyCommandCmdF() {
 		s.Len(printer.GetLines(), 1)
 		s.Len(printer.GetErrorLines(), 0)
 
-		check := getCommandFromCommandArg(s.th.SystemAdminClient, command.Id)
+		changedCommand, err := s.th.App.GetCommand(command.Id)
+		s.Require().Nil(err)
+		s.Require().Equal(url, changedCommand.URL)
 		s.Require().Equal(check.URL, url)
 	})
 
@@ -297,7 +299,7 @@ func (s *MmctlE2ETestSuite) TestModifyCommandCmdF() {
 		printer.Clean()
 		cmd := &cobra.Command{}
 
-		err := modifyCommandCmdF(s.th.SystemAdminClient, cmd, []string{"nothing"})
+		err := modifyCommandCmdF(c, cmd, []string{"nothing"})
 		s.Require().NotNil(err)
 		s.Require().Equal("unable to find command 'nothing'", err.Error())
 	})
@@ -314,7 +316,7 @@ func (s *MmctlE2ETestSuite) TestModifyCommandCmdF() {
 		})
 		s.Require().Nil(err)
 
-		err = modifyCommandCmdF(s.th.SystemAdminClient, cmd, []string{command.Id})
+		err = modifyCommandCmdF(c, cmd, []string{command.Id})
 		s.Require().NotNil(err)
 		s.Len(printer.GetLines(), 0)
 		s.Len(printer.GetErrorLines(), 0)
--- a/commands/command_e2e_test.go
+++ b/commands/command_e2e_test.go
@@ -265,7 +265,7 @@ func (s *MmctlE2ETestSuite) TestModifyCommandCmdF() {
 	command, _ := s.th.SystemAdminClient.CreateCommand(newCmd)
 	copy := command
 	originalURL := newCmd.URL
-
+	index := 0
 	s.RunForAllClients("modifyCommandCmdF", func(c client.Client) {
 		printer.Clean()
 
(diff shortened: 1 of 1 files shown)


commands/command_e2e_test.go
```diff
@@ -0,0 +1,73 @@
...
+		// Reset the cmd and parse to force Flag.Changed to be true.
+		cmd := CommandModifyCmd
+		cmd.ResetFlags()
+		addCommandFieldsFlags(cmd)
+		err := cmd.ParseFlags([]string{
+			command.Id,
+		})
+		s.Require().Nil(err)
+
+		err = modifyCommandCmdF(s.th.SystemAdminClient, cmd, []string{command.Id})
```
ashishbhate, line 40: This doesn't actually modify the command. I suggest modifying the command (maybe change the url) and then checking  that the command was actually modified. 
`RunForAllClients` runs the test function three times, once for each kind of client - System Admin, Local and User, so command should be reset between each test run.

commands/command_e2e_test.go
```diff
@@ -0,0 +1,73 @@
...
+		s.Require().Nil(err)
+		s.Len(printer.GetLines(), 1)
+		s.Len(printer.GetErrorLines(), 0)
+	})
+
+	s.RunForAllClients("modifyCommandCmdF with not existed command", func(c client.Client) {
```
ashishbhate, line 46:
```suggestion
	s.RunForAllClients("modifyCommandCmdF for command that does not exist", func(c client.Client) {
```

ashishbhate (commented):
Thanks @JeremyShih ! Looks good, some suggestions inline.

commands/command_e2e_test.go
JeremyShih: Updated to modify the url and then validated.
Also reset between each run

ashishbhate: @JeremyShih this only resets the command object here. It doesn't reset the command created on the server. After the first client run, the command on the server has the url set to `http://nowhere.com_modified`, and we can't be sure if the test runs for other clients will change it. 
I've added a suggestion for how we can get a unique url for each client's test run.

commands/command_e2e_test.go
```diff
@@ -0,0 +1,84 @@
...
+		printer.Clean()
+
+		command = copy
+		s.Require().Equal(command.URL, originalURL)
+
+		// Reset the cmd and parse to force Flag.Changed to be true.
+		cmd := CommandModifyCmd
+		cmd.ResetFlags()
+		addCommandFieldsFlags(cmd)
+		url := command.URL + "_modified"
```
ashishbhate, line 41:
```suggestion
	index := 0
	s.RunForAllClients("modifyCommandCmdF", func(c client.Client) {
		printer.Clean()

		command = copy
		s.Require().Equal(command.URL, originalURL)

		// Reset the cmd and parse to force Flag.Changed to be true.
		cmd := CommandModifyCmd
		cmd.ResetFlags()
		addCommandFieldsFlags(cmd)
		url := fmt.Sprintf("%s-%d", command.URL, index)
		index += 1
```

commands/command_e2e_test.go
JeremyShih: Thanks @ashishbhate!
I fixed some lint issue and it should be good.

The CircleCI Checks failed now and it says TestDeleteUsersCmd did not pass.
But I tried to run `make test-e2e` on local machine and it passed.
Could you help check with it?

commands/command_e2e_test.go
```diff
@@ -0,0 +1,87 @@
...
+		addCommandFieldsFlags(cmd)
+		url := fmt.Sprintf("%s-%d", command.URL, index)
+		index++
+		err := cmd.ParseFlags([]string{
+			command.Id,
+			"--url=" + url,
+		})
+		s.Require().Nil(err)
+
+		err = modifyCommandCmdF(s.th.SystemAdminClient, cmd, []string{command.Id})
```
mgdelacroix, line 51: For the command call, we should use the client provided by the `s.RunForAllClients` helper:

```suggestion
		err = modifyCommandCmdF(c, cmd, []string{command.Id})
```

commands/command_e2e_test.go
```diff
@@ -0,0 +1,87 @@
...
+		s.Require().Nil(err)
+		s.Len(printer.GetLines(), 1)
+		s.Len(printer.GetErrorLines(), 0)
+
+		check := getCommandFromCommandArg(s.th.SystemAdminClient, command.Id)
```
mgdelacroix, line 56: As here we're only checking the final state of the command, I'd suggest using the `th` to fetch the command directly through the app layer:

```suggestion
		changedCommand, err := s.th.App.GetCommand(command.Id)
		s.Require().Nil(err)
		s.Require().Equal(url, changedCommand.URL)
```

commands/command_e2e_test.go
```diff
@@ -0,0 +1,87 @@
...
+		s.Require().Equal(check.URL, url)
+	})
+
+	s.RunForAllClients("modifyCommandCmdF for command that does not exist", func(c client.Client) {
+		printer.Clean()
+		cmd := &cobra.Command{}
+
+		err := modifyCommandCmdF(s.th.SystemAdminClient, cmd, []string{"nothing"})
```
mgdelacroix, line 64: Same client suggestion here:

```suggestion
		err := modifyCommandCmdF(c, cmd, []string{"nothing"})
```

commands/command_e2e_test.go
```diff
@@ -0,0 +1,87 @@
...
+		cmd := CommandModifyCmd
+		cmd.ResetFlags()
+		addCommandFieldsFlags(cmd)
+		err := cmd.ParseFlags([]string{
+			command.Id,
+			"--trigger-word=modified with space",
+		})
+		s.Require().Nil(err)
+
+		err = modifyCommandCmdF(s.th.SystemAdminClient, cmd, []string{command.Id})
```
mgdelacroix, line 81: Same client suggestion here:

```suggestion
		err = modifyCommandCmdF(c, cmd, []string{command.Id})
```

mgdelacroix (commented):
Added come comments on top of @ashishbhate's ones

JeremyShih:
Applied and updated the codes.

commands/command_e2e_test.go
ashishbhate: @JeremyShih can you try merging master in your branch? I think your branch has diverged from master quite a bit. Let me know if you need help sorting out the merge conflict.

JeremyShih: It's good now after rebasing.

commands/command_e2e_test.go
```diff
@@ -249,3 +249,76 @@ func (s *MmctlE2ETestSuite) TestArchiveCommandCmdF() {
...
+
+	command, _ := s.th.SystemAdminClient.CreateCommand(newCmd)
+	copy := command
+	originalURL := newCmd.URL
+	index := 0
+	s.RunForSystemAdminAndLocal("modifyCommandCmdF", func(c client.Client) {
+		printer.Clean()
+
+		command = copy
+		s.Require().Equal(command.URL, originalURL)
```
ashishbhate, line 273: This assignment and check are not required. Since we're always changing the url, there's no need to check the original command object created. The variable `command` holds the original command created. The value stored in `command` is not the same as the command information stored on the server. 

Similarly we can remove lines 266 and 267.

JeremyShih: Updated

ashishbhate (approved):
Thanks @JeremyShih ! LGTM!

mgdelacroix (approved):
LGTM! Thanks @JeremyShih!

lindy65 (approved):
Thanks @JeremyShih @mgdelacroix - looks good to me 👍

PR #253, opened 2020-10-18
[MM-29566] Add e2e test for pluginMarketplaceInstallCmdF

wijayaerick:
#### Summary

Added e2e test for pluginMarketplaceInstallCmdF. 

#### Ticket Link

Fixes: mattermost/mattermost-server#15944

JIRA: https://mattermost.atlassian.net/browse/MM-29566

### DIFF
--- a/commands/plugin_marketplace_e2e_test.go
+++ b/commands/plugin_marketplace_e2e_test.go
@@ -1,0 +1,143 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package commands
+
+import (
+	"fmt"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/spf13/cobra"
+
+	"github.com/mattermost/mmctl/client"
+	"github.com/mattermost/mmctl/printer"
+)
+
+func (s *MmctlE2ETestSuite) TestPluginMarketplaceInstallCmd() {
+	s.SetupTestHelper().InitBasic()
+
+	s.RunForSystemAdminAndLocal("install a plugin", func(c client.Client) {
+		printer.Clean()
+
+		const (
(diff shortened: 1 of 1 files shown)


commands/plugin_marketplace_e2e_test.go
```diff
@@ -0,0 +1,143 @@
...
+			pluginID      = "jira"
+			pluginVersion = "invalid-version"
+		)
+
+		defer removePluginIfInstalled(s, pluginID)
+
+		err := pluginMarketplaceInstallCmdF(c, &cobra.Command{}, []string{pluginID, pluginVersion})
+		s.Require().NotNil(err)
+		s.Require().Contains(err.Error(), "Could not find the requested marketplace plugin.")
+		s.Require().Len(printer.GetErrorLines(), 0)
```
ashishbhate, line 107: @mgdelacroix its weird to me that `pluginMarketplaceInstallCmdF` just returns the error instead of printing it as well. Commands in `team.go`, `bot.go` and `user.go` print the error along with (or instead of) returning the error. 

Maybe we should fix this inconsistency?

ashishbhate (approved):
Thanks @wijayaerick ! LGTM.

lindy65 (approved):
Thanks @wijayaerick - looks good to me 👍

commands/plugin_marketplace_e2e_test.go
mgdelacroix: We're using both approaches depending on the case (sometimes if there is only one error for one operation, we just return it). One of the changes part of the `mmctl improvements` is to get rid of this inconsistency and basically print everything that is not a flag validation error, so I would suggest to tackle it then

mgdelacroix (approved):
LGTM! Thanks a lot @wijayaerick!! :confetti_ball:

PR #254, opened 2020-10-19
[MM-29335] end to end tests for delete channels

evilghostgirl:
#### Summary
Added tests to deleteChannelsCmdF. Tests includes a positive path, a path without permissions and a negative path -  when a non-existent resource are included.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15813


commands/channel_e2e_test.go
```diff
@@ -129,3 +129,95 @@ func (s *MmctlE2ETestSuite) TestUnarchiveChannelsCmdF() {
 		s.Require().Contains(printer.GetErrorLines()[0], "Unable to unarchive channel. The channel is not archived.")
 	})
 }
+
+func (s *MmctlE2ETestSuite) TestDeleteChannelsCmd() {
+	s.SetupTestHelper().InitBasic()
+
+	printer.SetFormat(printer.FormatPlain)
+	defer printer.SetFormat(printer.FormatJSON)
```
ashishbhate, line 137: This isn't strictly required.

commands/channel_e2e_test.go
```diff
@@ -129,3 +129,95 @@ func (s *MmctlE2ETestSuite) TestUnarchiveChannelsCmdF() {
...
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(fmt.Sprintf("Deleted channel '%s'", channel.Name), printer.GetLines()[0])
+		s.Require().Len(printer.GetErrorLines(), 0)
+
+		printer.Clean()
+		channel, _ = s.th.App.GetChannel(channel.Id)
+
+		s.Require().Nil(channel)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 0)
```
ashishbhate, line 178: The call to `s.th.App.GetChannel`  won't print any errors so we should check the error returned instead of the lines printed. 
See https://github.com/mattermost/mmctl/blob/master/commands/user_e2e_test.go#L311-L314 for an example.

commands/channel_e2e_test.go
```diff
@@ -129,3 +129,95 @@ func (s *MmctlE2ETestSuite) TestUnarchiveChannelsCmdF() {
...
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 1)
+		s.Require().Equal(fmt.Sprintf("Unable to find channel '%s:%s'", team.Id, otherChannel.Id), printer.GetErrorLines()[0])
+
+		printer.Clean()
+		channel, _ := s.th.App.GetChannel(otherChannel.Id)
+
+		s.Require().NotNil(channel)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 0)
```
ashishbhate, line 199: Same here, we don't need to check the printed lines.

commands/channel_e2e_test.go
```diff
@@ -129,3 +129,95 @@ func (s *MmctlE2ETestSuite) TestUnarchiveChannelsCmdF() {
...
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 1)
+		s.Require().Equal(fmt.Sprintf("Unable to find channel '%s:%s'", team.Id, notExistingChannelID), printer.GetErrorLines()[0])
+
+		printer.Clean()
+		channel, _ := s.th.App.GetChannel(notExistingChannelID)
+
+		s.Require().Nil(channel)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 0)
```
ashishbhate, line 221: Same here, we don't need to check the printed lines. Maybe we can check the error returned by `GetChannel` to confirm that the channel really doesn't exist.

.gitignore
```diff
@@ -1,7 +1,7 @@
 mmctl
 bin
 build
-
+.idea
```
ashishbhate, line 4: I dont think this change should be part of this PR, since its unrelated.

ashishbhate (commented):
Thanks @evilghostgirl ! Great effort, a few small comments inline.

commands/channel_e2e_test.go
evilghostgirl: When I delete above code then I getting error - in lines where printer is used. Something like that:
```
    Error:          Not equal: 
                                expected: string("Deleted channel 'channel_name'")
                                actual  : *model.Channel(&model.Channel{Id:"zzgmgm8snbd8xkc3x7s1fp5mnr", CreateAt:1603220156368, UpdateAt:1603220156368, DeleteAt:0, TeamId:"", Type:"O", DisplayName:"", Name:"channel_name", Header:"", Purpose:"", LastPostAt:0, TotalMsgCount:0, ExtraUpdateAt:0, CreatorId:"a6n39sfb9tdsz8qpbcinmy4k5a", SchemeId:(*string)(nil), Props:map[string]interface {}(nil), GroupConstrained:(*bool)(nil)})

```
Theoretically, I can delete every line where printer is used - because it is unit tested, but maybe it would be better to check if the output works properly.

ashishbhate: `printer.GetLines` returns `printer.Lines` which is just a `[]interface{}`. With `printer.FormatJSON` the struct passed to the printer is stored in `printer.Lines` directly.

Its a bit unintuitive, but you can just compare the object directly.

So instead of `s.Require().Equal(fmt.Sprintf("Deleted channel '%s'", channel.Name), printer.GetLines()[0])`

you can write: `s.Require().Equal(channel, printer.GetLines()[0])`

This isn't a blocker, the way you've written it is fine as well.

commands/channel_e2e_test.go
```diff
@@ -129,3 +129,93 @@ func (s *MmctlE2ETestSuite) TestUnarchiveChannelsCmdF() {
...
+
+		printer.Clean()
+		err := deleteChannelsCmdF(c, cmd, args)
+
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(fmt.Sprintf("Deleted channel '%s'", channel.Name), printer.GetLines()[0])
+		s.Require().Len(printer.GetErrorLines(), 0)
+
+		printer.Clean()
```
ashishbhate, line 173: this call is not required.

commands/channel_e2e_test.go
```diff
@@ -129,3 +129,93 @@ func (s *MmctlE2ETestSuite) TestUnarchiveChannelsCmdF() {
...
+		err := deleteChannelsCmdF(c, cmd, args)
+
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 1)
+		s.Require().Equal(fmt.Sprintf("Unable to find channel '%s:%s'", team.Id, notExistingChannelID), printer.GetErrorLines()[0])
+
+		printer.Clean()
```
ashishbhate, line 214: Similarly, this call is not required.

ashishbhate (commented):
Just a few small comments inline.

ashishbhate (approved):
LGTM!

lindy65 (approved):
Thanks @evilghostgirl - looks good to me 👍

mgdelacroix (approved):
LGTM! Thanks for the contribution @evilghostgirl!! :confetti_ball:

PR #278, opened 2020-11-04
[MM-30264] Increase test timeout

mgdelacroix:
#### Summary
Increases the timeout of the tests to avoid CI failures

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30264


Makefile
```diff
@@ -85,21 +85,21 @@ test: test-unit
 .PHONY: test-unit
 test-unit:
 	@echo Running unit tests
-	$(GO) test -mod=vendor -race -v -tags unit $(GO_PACKAGES)
+	$(GO) test -mod=vendor -timeout 30m -race -v -tags unit $(GO_PACKAGES)
```
hanzei, line 88: Is it worth extracting these flags into a a separate variable to DRY the code?

hanzei (approved):
One non blocking question. LGTM :+1:

Makefile
mgdelacroix: Make sense, added in bd8425b

jespino (approved):
LGTM

PR #283, opened 2020-11-16
MM-28227: Adds command to reset default role permissions.

mkraft:
#### Summary

Add new command `mmctl permissions reset <role_name>` to reset the permissions to their defaults.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28227


hahmadia (commented):
Code looks good but I had one thought.
Would it make sense to implement this on the API side rather than in the command itself so that if we want to implement a UI for it, it be easier to do and we don't have to duplicate the logic? From the command you can use the client and in the UI we can just directly call the endpoint. Something like `roles/{role.id}/reset`.

Any thoughts or comments on this?

mkraft:
> Code looks good but I had one thought.
> Would it make sense to implement this on the API side rather than in the command itself so that if we want to implement a UI for it, it be easier to do and we don't have to duplicate the logic? From the command you can use the client and in the UI we can just directly call the endpoint. Something like `roles/{role.id}/reset`.
> 
> I can see in the future we may want a reset role similar to how we have reset permissions in the system permission scheme. That's where the above idea came from.
> 
> Any thoughts or comments on this?

I think it's a good consideration. I think if the logic weren't so trivial—just retrieving a slice from a map—then it might be worth doing it preemptively (before we need the API). I'd rather wait until there's some need for it on the API side to avoid the extra overhead of dev., QA, and maintenance.

hahmadia:
Explained to me in meeting. Thanks!

hahmadia (approved):

commands/permissions.go
```diff
@@ -124,6 +137,28 @@ func removePermissionsCmdF(c client.Client, cmd *cobra.Command, args []string) e
...
+	defaultRole, ok := model.MakeDefaultRoles()[role.Name]
+	if !ok {
+		return fmt.Errorf("no default permissions available for role")
+	}
+
+	patchRole := model.RolePatch{
+		Permissions: &defaultRole.Permissions,
+	}
+
+	if _, response = c.PatchRole(role.Id, &patchRole); response.Error != nil {
```
mgdelacroix, line 155: We should print the returned role if the reset is successful

commands/permissions_test.go
```diff
@@ -187,3 +187,70 @@ func (s *MmctlUnitTestSuite) TestRemovePermissionsCmd() {
 		s.Require().EqualError(err, "Role: role_not_found, ")
 	})
 }
+
+func (s *MmctlUnitTestSuite) TestResetPermissionsCmd() {
```
mgdelacroix, line 191: Can you please add a couple of end to end tests to the PR?

mkraft: > Could you please add a couple of end to end tests to the PR?

What's the intent you're thinking? To ensure that `PatchRole` does what it should?

mgdelacroix: Mainly the communication between the tool and the server, yes. All the permutation tests like args, errors and such should be tested in the unit test layer, and for the end to end we're trying to add a happy path case to ensure that we're not missing anything with the mocks and a permissions error case if there is any.

On top of those we can add more cases if they are particularly interesting to test for a given command, but if not these two are enough.

commands/permissions.go
```diff
@@ -46,11 +49,22 @@ var ShowRoleCmd = &cobra.Command{
 	RunE:       withClient(showRoleCmdF),
 }
 
+var ResetCmd = &cobra.Command{
+	Use:   "reset <role_name>",
+	Short: "Reset default permissions for role (EE Only)",
+	Long:  "Reset the given role's permissions to the set that was originally released with",
+	Example: `  # Reset the permissions of the 'system_read_only_admin' role.
+  permissions reset system_read_only_admin`,
```
mgdelacroix, line 57: small, nonblocking nit I missed before:
```suggestion
  $ mmctl permissions reset system_read_only_admin`,
```

commands/permissions_reset_e2e_test.go
```diff
@@ -0,0 +1,78 @@
...
+		s.Require().Error(err2)
+		s.Require().Len(printer.GetLines(), 0)
+		s.Require().Len(printer.GetErrorLines(), 0)
+
+		// ensure reset didn't happen
+		roleAfterResetAttempt, err := s.th.App.GetRoleByName(model.SYSTEM_USER_MANAGER_ROLE_ID)
+		s.Require().Nil(err)
+		sort.Strings(expectedPermissions)
+		sort.Strings(roleAfterResetAttempt.Permissions)
+		s.Require().Equal(expectedPermissions, roleAfterResetAttempt.Permissions)
```
mgdelacroix, line 44: nit: this can be simplified with [ElementsMatch](https://godoc.org/github.com/stretchr/testify/require#ElementsMatch)

```suggestion
		s.Require().ElementsMatch(expectedPermissions, roleAfterResetAttempt.Permissions)
```

commands/permissions_reset_e2e_test.go
```diff
@@ -0,0 +1,78 @@
...
+		s.Require().Nil(err2)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Len(printer.GetErrorLines(), 0)
+
+		// ensure reset was successful
+		roleAfterResetAttempt, err := s.th.App.GetRoleByName(model.SYSTEM_USER_MANAGER_ROLE_ID)
+		s.Require().Nil(err)
+		sort.Strings(defaultPermissions)
+		sort.Strings(roleAfterResetAttempt.Permissions)
+		s.Require().Equal(defaultPermissions, roleAfterResetAttempt.Permissions)
```
mgdelacroix, line 76: Same:

```suggestion
		s.Require().ElementsMatch(defaultPermissions, roleAfterResetAttempt.Permissions)
```

mgdelacroix (approved):
Thanks for applying the changes @mkraft!! LGTM! :+1: 

Added a couple of nonblocking nits

lindy65 (approved):
Thanks @mkraft - looks good to me 👍 

I see there is still a PM: Review label on the PR - does a PM still need to review?

PR #320, opened 2021-02-15
Use newer golangci-lint version

metanerd:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,7 +20,7 @@ docs:
 
 lint:
   stage: test
-  image: $CI_REGISTRY/mattermost/ci/images/golangci-lint:v1.25.1-1
+  image: $CI_REGISTRY/mattermost/ci/images/golangci-lint:v1.31.0-1
   script:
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Thanks :+1:

streamer45:
Not sure why this is consistently failing several tests. @mgdelacroix do you have a clue?

mgdelacroix:
@streamer45 weird behavior indeed, the image is constrained to the `lint` job, so it should not affect to the `test` one, which is actually passing in `master` :thinking:

streamer45:
> @streamer45 weird behavior indeed, the image is constrained to the `lint` job, so it should not affect to the `test` one, which is actually passing in `master`

Oh well, this is getting merged into my PR not into master. Had to deal with a few unrelated changes (breaking of client API due to collapsed threads and Golang version bump) so probably something got broken. I can merge this and try to deal with the issues there I guess.

mgdelacroix:
Totally missed the fact that this was not being migrated on `master` :grimacing: makes sense to deal with the issue in your branch then :)

PR #321, opened 2021-02-22
feat(commands): added displaying user roles on update

parsaakbari1209:
#### Summary

Display user roles on update.
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32395

### DIFF
--- a/commands/roles.go
+++ b/commands/roles.go
@@ -72,6 +72,8 @@ func rolesSystemAdminCmdF(c client.Client, _ *cobra.Command, args []string) erro
 
 			printer.Print(fmt.Sprintf("System admin role assigned to user %q", args[i]))
 		}
+
+		printer.Print(fmt.Sprintf("Updated %s roles: %s", user.Email, user.Roles))
 	}
 
 	return nil
@@ -106,6 +108,8 @@ func rolesMemberCmdF(c client.Client, _ *cobra.Command, args []string) error {
 
 			printer.Print(fmt.Sprintf("System admin role revoked for user %q", args[i]))
 		}
+
+		printer.Print(fmt.Sprintf("Updated %s roles: %s", user.Email, user.Roles))
 	}
 
 	return nil
--- a/commands/roles.go
+++ b/commands/roles.go
@@ -70,10 +70,8 @@ func rolesSystemAdminCmdF(c client.Client, _ *cobra.Command, args []string) erro
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


commands/roles.go
```diff
@@ -72,6 +72,8 @@ func rolesSystemAdminCmdF(c client.Client, _ *cobra.Command, args []string) erro
 
 			printer.Print(fmt.Sprintf("System admin role assigned to user %q", args[i]))
 		}
+
+		printer.Print(fmt.Sprintf("Updated %s roles: %s", user.Email, user.Roles))
```
mgdelacroix, line 76: The roles that `user` contains are not updated with the new information that resulted from the `c.UpdateUserRoles`, so this will only print whatever roles the user had before making that request.

Instead, we could use the roles in the `roles` list that we have built before, assuming that if the request succeeded, the user will have them applied.

One last point would be that we should print this information as part of the `printer.Print` call in line 73, as this will only get printed if the request succeeds. So maybe something like:

```suggestion
			printer.Print(fmt.Sprintf("System admin role assigned to user %q. Current roles are: %s", args[i], strings.Join(", ", roles))
		}
```

commands/roles.go
```diff
@@ -106,6 +108,8 @@ func rolesMemberCmdF(c client.Client, _ *cobra.Command, args []string) error {
 
 			printer.Print(fmt.Sprintf("System admin role revoked for user %q", args[i]))
 		}
+
+		printer.Print(fmt.Sprintf("Updated %s roles: %s", user.Email, user.Roles))
```
mgdelacroix, line 112: Here we should do something similar, merging the print with the one inside the if condition and printing the roles in `newRoles` instead of the `user` ones.

mgdelacroix (changes requested):
Thanks for the contribution @parsaakbari1209!! Some changes requested

parsaakbari1209:
Yeah, thank you for the awesome review, I'll change them right away.
It is kind of my first open source contribution 🔥

mgdelacroix (approved):
Thanks a lot for applying the changes @parsaakbari1209!! Looks great!! :rocket:

isacikgoz (approved):
Awesome work @parsaakbari1209, LGTM 🎉

lindy65 (approved):
Thanks @parsaakbari1209 - apologies that your PR has been waiting to be QA-reviewed! I've added myself as reviewer and this will be tested after merge 👍 

//cc @mgdelacroix @isacikgoz

mgdelacroix (approved):

PR #342, opened 2021-03-17
[MM-32044] Reset SAML auth data

maxerenberg:
#### Summary
This PR adds an mmctl command that reset SAML users' AuthData field to their email.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32044

#### Server PR
https://github.com/mattermost/mattermost-server/pull/17161

### DIFF
--- a/commands/saml.go
+++ b/commands/saml.go
@@ -22,13 +22,13 @@ var SamlAuthDataResetCmd = &cobra.Command{
 	Short: "Reset AuthData field to Email",
 	Long:  "Resets the AuthData field for SAML users to their email. Run this utility after setting the 'id' SAML attribute to an empty value.",
 	Example: `  # Reset all SAML users' AuthData field to their email, including deleted users
-$ mmctl saml reset-auth-data --include-deleted
+  $ mmctl saml reset-auth-data --include-deleted
 
-# Show how many users would be affected by the reset
-$ mmctl saml reset-auth-data --dry-run
+  # Show how many users would be affected by the reset
+  $ mmctl saml reset-auth-data --dry-run
 
-# Only reset the AuthData for the following SAML users
-$ mmctl saml reset-auth-data --users userid1,userid2`,
+  # Only reset the AuthData for the following SAML users
+  $ mmctl saml reset-auth-data --users userid1,userid2`,
 	RunE: withClient(samlAuthDataResetCmdF),
 }
 
--- a/commands/saml.go
+++ b/commands/saml.go
@@ -17,20 +17,27 @@ var SamlCmd = &cobra.Command{
 	Short: "SAML related utilities",
 }
 
-var SamlAuthDataReset = &cobra.Command{
-	Use:     "authdatareset",
-	Short:   "Reset AuthData field to Email",
-	Long:    "Resets the AuthData field for SAML users to their email. Run this utility after setting the 'id' SAML attribute to an empty value.",
-	Example: " saml authdatamigrate",
-	RunE:    withClient(samlAuthDataResetCmdF),
+var SamlAuthDataResetCmd = &cobra.Command{
+	Use:   "auth-data-reset",
+	Short: "Reset AuthData field to Email",
+	Long:  "Resets the AuthData field for SAML users to their email. Run this utility after setting the 'id' SAML attribute to an empty value.",
+	Example: `  # Reset all SAML users' AuthData field to their email, including deleted users
+$ mmctl saml reset-auth-data --include-deleted
+
+# Show how many users would be affected by the reset
+$ mmctl saml reset-auth-data --dry-run
+
+# Only reset the AuthData for the following SAML users
+$ mmctl saml reset-auth-data --users userid1,userid2`,
+	RunE: withClient(samlAuthDataResetCmdF),
 }
 
 func init() {
-	SamlAuthDataReset.Flags().Bool("include-deleted", false, "Include deleted users")
-	SamlAuthDataReset.Flags().Bool("dry-run", false, "Dry run only")
-	SamlAuthDataReset.Flags().StringSlice("users", nil, "Comma-separated list of user IDs to which the operation will be applied")
+	SamlAuthDataResetCmd.Flags().Bool("include-deleted", false, "Include deleted users")
+	SamlAuthDataResetCmd.Flags().Bool("dry-run", false, "Dry run only")
+	SamlAuthDataResetCmd.Flags().StringSlice("users", nil, "Comma-separated list of user IDs to which the operation will be applied")
 	SamlCmd.AddCommand(
-		SamlAuthDataReset,
+		SamlAuthDataResetCmd,
 	)
 	RootCmd.AddCommand(SamlCmd)
 }
--- a/commands/saml.go
+++ b/commands/saml.go
@@ -36,6 +36,7 @@ func init() {
 	SamlAuthDataResetCmd.Flags().Bool("include-deleted", false, "Include deleted users")
 	SamlAuthDataResetCmd.Flags().Bool("dry-run", false, "Dry run only")
 	SamlAuthDataResetCmd.Flags().StringSlice("users", nil, "Comma-separated list of user IDs to which the operation will be applied")
+
 	SamlCmd.AddCommand(
 		SamlAuthDataResetCmd,
 	)
--- a/commands/saml.go
+++ b/commands/saml.go
@@ -4,7 +4,9 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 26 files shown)


commands/saml.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var SamlCmd = &cobra.Command{
+	Use:   "saml",
+	Short: "SAML related utilities",
+}
+
+var SamlAuthDataReset = &cobra.Command{
```
mgdelacroix, line 20: For consistency this should be `SamlAuthDataResetCmd`

commands/saml.go
```diff
@@ -0,0 +1,55 @@
...
+	Use:     "authdatareset",
```
mgdelacroix, line 21: What about using `auth-data-reset` here instead for readability?

commands/saml.go
```diff
@@ -0,0 +1,55 @@
...
+	Short:   "Reset AuthData field to Email",
+	Long:    "Resets the AuthData field for SAML users to their email. Run this utility after setting the 'id' SAML attribute to an empty value.",
+	Example: " saml authdatamigrate",
```
mgdelacroix, line 24: Can you please use the new example format to elaborate on what things can be done with the command? A good example is the `mmctl user create` command

commands/saml.go
```diff
@@ -0,0 +1,55 @@
...
+	RunE:    withClient(samlAuthDataResetCmdF),
+}
+
+func init() {
+	SamlAuthDataReset.Flags().Bool("include-deleted", false, "Include deleted users")
+	SamlAuthDataReset.Flags().Bool("dry-run", false, "Dry run only")
+	SamlAuthDataReset.Flags().StringSlice("users", nil, "Comma-separated list of user IDs to which the operation will be applied")
```
mgdelacroix, line 31: If this flag is required, we can use something like `SamlAuthDataReset.MarkFlagRequired("users")` here, tho I don't know if this works with `StringSlice`. If it doesn't, we can check that `len(userIDs) > 0` inside the command

commands/saml_test.go
```diff
@@ -0,0 +1,48 @@
...
+			EXPECT().
+			ResetSamlAuthDataToEmail(false, true, []string{}).
+			Return(int64(1), &model.Response{Error: nil})
+
+		err := samlAuthDataResetCmdF(s.client, cmd, nil)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0], outputMessage)
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
```
mgdelacroix, line 47: Can we add a test case to test that the user list flag gets passed to the request?

commands/saml_test.go
```diff
@@ -0,0 +1,48 @@
...
+}
```
mgdelacroix, line 48: Can you please add some e2e tests as well?

mgdelacroix (commented):
Looking good! Some comments added related to command interactive docs and tests

commands/saml.go
maxerenberg: The flag is not required. The backend will handle the cases where userIDs is null or empty.

mgdelacroix: I see, makes sense, then I think this is a good case for using the `examples` to let the user know what will happen if the flag is not set :)

commands/saml_test.go
maxerenberg: E2E tests will be in another PR since https://github.com/mattermost/mmctl/pull/310 has not been merged yet.

commands/saml.go
```diff
@@ -0,0 +1,62 @@
...
+# Only reset the AuthData for the following SAML users
+$ mmctl saml reset-auth-data --users userid1,userid2`,
+	RunE: withClient(samlAuthDataResetCmdF),
+}
+
+func init() {
+	SamlAuthDataResetCmd.Flags().Bool("include-deleted", false, "Include deleted users")
+	SamlAuthDataResetCmd.Flags().Bool("dry-run", false, "Dry run only")
+	SamlAuthDataResetCmd.Flags().StringSlice("users", nil, "Comma-separated list of user IDs to which the operation will be applied")
+	SamlCmd.AddCommand(
```
mgdelacroix, line 39:
```suggestion

	SamlCmd.AddCommand(
```

commands/saml_test.go
```diff
@@ -0,0 +1,63 @@
...
+			ResetSamlAuthDataToEmail(false, false, []string{}).
+			Return(int64(1), &model.Response{Error: nil})
+
+		err := samlAuthDataResetCmdF(s.client, &cobra.Command{}, nil)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0], outputMessage)
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+	s.Run("Reset auth data dry run", func() {
```
mgdelacroix, line 30:
```suggestion

	s.Run("Reset auth data dry run", func() {
```

commands/saml_test.go
```diff
@@ -0,0 +1,63 @@
...
+		printer.Clean()
+		outputMessage := "1 user records would be affected.\n"
+
+		cmd := &cobra.Command{}
+		cmd.Flags().Bool("dry-run", true, "")
+
+		s.client.
+			EXPECT().
+			ResetSamlAuthDataToEmail(false, true, []string{}).
+			Return(int64(1), &model.Response{Error: nil})
```
mgdelacroix, line 40:
```suggestion
			Return(int64(1), &model.Response{Error: nil}).
			Times(1)
```

commands/saml_test.go
```diff
@@ -0,0 +1,63 @@
...
+		s.Require().Equal(printer.GetLines()[0], outputMessage)
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+	s.Run("Reset auth data with specified users", func() {
+		printer.Clean()
+		users := []string{"user1"}
+		s.client.
+			EXPECT().
+			ResetSamlAuthDataToEmail(false, false, users).
+			Return(int64(1), &model.Response{Error: nil})
```
mgdelacroix, line 54:
```suggestion
			Return(int64(1), &model.Response{Error: nil}).
			Times(1)
```

commands/saml_test.go
```diff
@@ -0,0 +1,63 @@
...
+			ResetSamlAuthDataToEmail(false, true, []string{}).
+			Return(int64(1), &model.Response{Error: nil})
+
+		err := samlAuthDataResetCmdF(s.client, cmd, nil)
+		s.Require().Nil(err)
+		s.Require().Len(printer.GetLines(), 1)
+		s.Require().Equal(printer.GetLines()[0], outputMessage)
+		s.Require().Len(printer.GetErrorLines(), 0)
+	})
+	s.Run("Reset auth data with specified users", func() {
```
mgdelacroix, line 48:
```suggestion

	s.Run("Reset auth data with specified users", func() {
```

commands/saml_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (s *MmctlUnitTestSuite) TestSamlAuthDataReset() {
+	s.Run("Reset auth data without errors", func() {
+		printer.Clean()
+		outputMessage := "1 user records were changed.\n"
+
+		s.client.
+			EXPECT().
+			ResetSamlAuthDataToEmail(false, false, []string{}).
+			Return(int64(1), &model.Response{Error: nil})
```
mgdelacroix, line 22: nit: adding `Times` call to keep consistency with the rest of the tests

```suggestion
			Return(int64(1), &model.Response{Error: nil}).
			Times(1)
```

mgdelacroix (commented):
Thanks for applying the changes!! Made some nit comments related to styling and consistency, and I think the users flag test is still missing, but otherwise LGTM

maxerenberg:
> I think the users flag test is still missing
I believe this is case is accounted for in the "Reset auth data with specified users" test?

mgdelacroix (approved):
Right, missed that test case, LGTM! :+1:

commands/saml_test.go
maxerenberg: E2E tests are here: https://github.com/mattermost/mmctl/pull/345

sbishel (commented):
Similar to API, lets change `dry-run` to `execute`.

sbishel (approved):

furqanmlk (approved):
LGTM !
Added tests details here
https://github.com/mattermost/mattermost-server/pull/17161

PR #349, opened 2021-04-15
MM-34875: Update Mattermost dependency to latest

agnivade:
https://mattermost.atlassian.net/browse/MM-34875

This also fixes some tests which were changed because of product behavior:
- Job not found error due to mattermost/mattermost-server#16917.
- MFA disabled error due to mattermost/mattermost-server#16994.
- Minor system console printing format changes.


agnivade:
Oh dear, several tests are failing. :(

agnivade:
So it seems like one of the broken tests is due to https://github.com/mattermost/mattermost-server/pull/16917. @hahmadia - is it intentional that a GetJob with ID will return an "Unable to get the job." error now instead of the "You do not have the appropriate permissions., " error?

@mgdelacroix - I am wondering whether we should run nightly tests with MM master branch and mmctl? This should help us catch failures more quickly rather than wait for a full release, and then fix a lot of issues. Right now, as it stands, several tests are failing, and someone needs to go through each of them, and track down to understand why it's failing and whether it is mmctl's fault or MM's fault.

mgdelacroix:
@agnivade traditionally we've done this integrating the tests of the external repo into `mattermost-server` to block changes that would affect them (think of `enterprise`) but back in the day the tests would take too long already and they would not trigger that many failures in `mmctl`. Maybe now we can add a job on the CI, or if we deem that the time/benefit is not that high, I like a lot the idea of running nightly tests and it's a lot more cost effective.

agnivade:
Sounds good. Probably we'll need to talk with the release team to add something like this.

hahmadia:
@agnivade 
I found the test you were talking about.
https://github.com/mattermost/mmctl/blob/master/commands/export_e2e_test.go#L321

After the changes you linked to were made, the job is retrieved first and then the job type is used to determine if we have permissions or not to creating that job type. 

Since a invalid job ID is being passed in that test, it tries to retrieve the job which it can't and throws the error you mentioned so it is working as expected.

If we still want it to show the no permissions error as it currently expects, we need to change the test to something like 
https://github.com/mattermost/mmctl/blob/master/commands/export_e2e_test.go#L339 where we save a job first and pass it a valid job id.

agnivade:
Yes, we can change the test. I just wanted confirmation that showing "unable to get the job" instead of a permission error for an invalid job id is indeed an intentional change, and not an oversight. Because the PR seemed to do something else.

From your response, it sounds like intentional, so I will change the test.

agnivade:
@mgdelacroix - This is good to be reviewed now.

mgdelacroix (approved):
LGTM! :+1: Thanks for taking care of the test fixes!

PR #356, opened 2021-05-13
Update the ldap idmigrate docs to add the command's positional argument

mgdelacroix:


cwarnermm (approved):
Great update! Thanks!

PR #375, opened 2021-08-09
[MM-37687] Add pagination to channel list command

mgdelacroix:
#### Summary
This PR adds a loop for each of the channel fetch functions, so the `channel list` command doesn't hit the max per page limit in the server and effectively fetches all the channels available.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37687

### DIFF
--- a/commands/channel.go
+++ b/commands/channel.go
@@ -11,6 +11,8 @@ import (
 	"github.com/mattermost/mmctl/printer"
 
 	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/mattermost/mattermost-server/v6/web"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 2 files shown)


lindy65 (approved):
Thanks @mgdelacroix - this will be tested after merge 👍

commands/channel.go
```diff
@@ -263,6 +265,48 @@ func archiveChannelsCmdF(c client.Client, cmd *cobra.Command, args []string) err
...
+	channels := []*model.Channel{}
+	page := 0
+
+	for {
+		channelsPage, response := c.GetPublicChannelsForTeam(teamID, page, web.PerPageMaximum, "")
+		if response.Error != nil {
+			return nil, response.Error
+		}
+
+		if len(channelsPage) == 0 {
```
larkox, line 278: Probably not needed, but should we check later also `len(channelsPage) < web.PerPageMaximum` to save us from one more call?

larkox (approved):
LGTM. Just two small comments.

commands/channel.go
```diff
@@ -271,25 +315,25 @@ func listChannelsCmdF(c client.Client, cmd *cobra.Command, args []string) error
 			continue
 		}
 
-		publicChannels, response := c.GetPublicChannelsForTeam(team.Id, 0, 10000, "")
-		if response.Error != nil {
-			printer.PrintError("Unable to list public channels for '" + args[i] + "'. Error: " + response.Error.Error())
+		publicChannels, err := getAllPublicChannelsForTeam(c, team.Id)
+		if err != nil {
+			printer.PrintError(fmt.Sprintf("unable to list public channels for %q: %s", args[i], err))
```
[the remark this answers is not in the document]
mgdelacroix: The change was intended, just changing for consistency

commands/channel.go
mgdelacroix: We've found instances in which although the page comes with less than the total amount of elements, there are more pages to fetch (for example, when the items are indexed on ES/bleve) hence we stop only when the page fetched comes empty, to be 100% sure that there are no more elements to fetch

isacikgoz (approved):
LGTM 👍

PR #384, opened 2021-08-24
Use consistent image

metanerd:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ executors:
   default:
     working_directory: ~/go/src/github.com/mattermost/mmctl
     docker:
-      - image: circleci/golang:1.15.5
+      - image: circleci/golang:1.17.0
 
 aliases:
 - &restore_cache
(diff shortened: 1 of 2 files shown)


metanerd:
Breaking build step is known to the introducer: https://github.com/mattermost/mmctl/commits/master
I think this could be merged already, even before the build is fixed.

.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ commands:
     parameters:
       version:
         type: string
-        default: 1.31.0
+        default: 1.42.0
```
streamer45, line 28: Is this needed or just updating while we are here?

metanerd: Just updating while we are here.

streamer45: Alright, as long as it doesn't cause new errors we should be good :)

streamer45 (approved):

.circleci/config.yml
metanerd: You are right, might need a cache key bump.

PR #402, opened 2021-10-29
add arm64 targets and update mm deps to 6.0.2

cpanato:
#### Summary
add arm64 targets and update mm deps to 6.0.2


mgdelacroix (approved):
LGTM!! :+1:

PR #418, opened 2021-12-13
[MM-40604] Enable parsing of floating point values from config

streamer45:
#### Summary

Numeric values coming from plugins' config are mapped to floating point numbers. This made it impossible to update them from `mmctl`.

E.g.

`mmctl --local config set PluginSettings.Plugins.com.mattermost.calls.udpserverport 8443`

would fail with:

```
Error: target value type is not supported
exit status 1
```

PR adds support for such conversion.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40604

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -208,6 +208,14 @@ func setValueWithConversion(val reflect.Value, newValue interface{}) error {
 		}
 		val.SetInt(v)
 		return nil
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
LGTM 👍

mgdelacroix (approved):
LGTM! :+1:

PR #445, opened 2022-02-02
Pin golangci to a version

agnivade:
Master is broken due to this. Pinning this keeps the code stable and we can make changes slowly.


cpanato (approved):
golangci-lint is on version v1.44.0, it is okay we keep this old version for now

cpanato:
will do manually

PR #518, opened 2022-05-17
MM-38703: Show error detail while printing jobs

agnivade:
We show the job metadata which captures detail on why
a job has failed. This can be helpful while debugging.

https://mattermost.atlassian.net/browse/MM-38703

```release-note
NONE
```


isacikgoz (approved):
LGTM 👍

PR #536, opened 2022-07-19
Makefile sets DIST_VER value empty when there is no tag.

pfltdv:
#### Summary
PR contains changes to check DIST_VER variable with undefined/empty values. When it is empty it will set to latest tag available.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-3718


phoinixgrr (approved):
🎖️

PR #553, opened 2022-09-02
Add the option to disable newlines and pre-parse templates

noxer:
#### Summary
In order to present the progress of the import file validation, we need a way to render lines without a newline. Also due to the progress being rendered multiple times per second (on a half decent machine), this PR adds the possibility to pass pre-parsed templates to the printer to avoid overhead.

#### Ticket Link
[MM-38705](https://mattermost.atlassian.net/browse/MM-38705)

### DIFF
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -26,12 +26,14 @@ type Printer struct { //nolint
 	writer  io.Writer
 	eWriter io.Writer
 
-	Format     string
-	Single     bool
-	pager      bool
-	Quiet      bool
-	Lines      []interface{}
-	ErrorLines []interface{}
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):

agnivade (commented):
Add a test?

Also curious to know how it replaces the same line? I don't see a curses library being used.

noxer:
[x] Test added
It replaces the same line by printing a `\r` to return the text cursor at the start of the current line without a line break. So the next line will replace the old one.

agnivade:
Thanks, I was thinking something like that.

PR #609, opened 2022-11-11
MM-46957 Use the server import validation code in mmctl

noxer:
#### Summary
In order to avoid discrepancies between `mmctl import validate` and the actual server import this PR merges the code between them. The `mmctl` validation code was a copy of the server code in the first place, so the merge is not too difficult.
To make this work I had to update the server dependency to 7.4.0 (because we are still using an incompatible versioning scheme, the version listed is 6.0.0, while the commit points to the 7.4.0 release commit).

The biggest change is in the vendor folder, also the e2e tests had to to be adapted to changes in the errors returned by the server.
You can find the actual changes limited to the `importer/` folder.

#### Ticket Link
[MM-46957](https://mattermost.atlassian.net/browse/MM-46957)


isacikgoz:
image

PR #629, opened 2023-01-13
Remove --local option from bot create doc since it is not available

m1lt0n:
#### Summary

The documentation on `bot create` lists `--local` as an option, but it is disabled/not available for the specific command. This PR removes `--local` from the list of available options.

### DIFF
--- a/docs/mmctl_bot_create.rst
+++ b/docs/mmctl_bot_create.rst
@@ -42,7 +42,6 @@ Options inherited from parent commands
       --insecure-sha1-intermediate   allows to use insecure TLS protocols, such as SHA-1
       --insecure-tls-version         allows to use TLS versions 1.0 and 1.1
       --json                         the output format will be in json format
-      --local                        allows communicating with the server through a unix socket
       --quiet                        prevent mmctl to generate output for the commands
       --strict                       will only run commands if the mmctl version matches the server one
       --suppress-warnings            disables printing warning messages


cwarnermm (approved):
Thanks, @m1lt0n! 

@mgdelacroix - @m1lt0n has also updated the product documentation with this update via https://github.com/mattermost/docs/pull/6153

m1lt0n:
@cwarnermm the docs here should be autogenerated (as the failing CI step mentions) and this is the reason why here all commands have the same flags/options (as they are the same for all commands), so this is a bit tricky (the bot create command is the only exception that disallows the `local` option, also). So this needs some thought as to if it's possible to have different set of options for different commands.

cwarnermm:
You're absolutely right about these docs being generated, @m1lt0n. I apologize for sending you down this path in error. 

@mgdelacroix - Can you advise on next steps for this mmctl generated docs update?

mgdelacroix:
@cwarnermm the `--local` flag is global to all commands and inherited by them, so it's tricky to remove from just one of them. I've been playing with the idea of overwriting the help function so the flag is not shown, something like:

```diff
diff --git a/commands/bot.go b/commands/bot.go
index d502a8bf..3d6771c6 100644
--- a/commands/bot.go
+++ b/commands/bot.go
@@ -80,6 +80,10 @@ func init() {
        CreateBotCmd.Flags().String("display-name", "", "Optional. The display name for the new bot.")
        CreateBotCmd.Flags().String("description", "", "Optional. The description text for the new bot.")
        CreateBotCmd.Flags().Bool("with-token", false, "Optional. Auto genreate access token for the bot.")
+       CreateBotCmd.SetHelpFunc(func(c *cobra.Command, args []string) {
+               c.Flags().MarkHidden("local")
+               c.Parent().HelpFunc()(c, args)
+       })
        ListBotCmd.Flags().Bool("orphaned", false, "Optional. Only show orphaned bots.")
        ListBotCmd.Flags().Bool("all", false, "Optional. Show all bots (including deleleted and orphaned).")
        UpdateBotCmd.Flags().String("username", "", "Optional. The new username for the bot.")
```

but although this will fix the output of `mmctl bot create --help`, the flag will still be included in the files generated by `make docs`. My suggestion would be to leave it as is and let the command handle the error when it cannot be called with the `--local` flag enabled. The flag itself is listed under the "inherited from parent commands" section, so it's already addressed as non-specific to the command, and could behave as such

m1lt0n:
@cwarnermm In that case I suggest we close this PR, but I believe it's good that the documentation change that has been merge does not mention local as a valid/available option, so despite it looks like a small inconsistency, I think it's ok.

cc: @mgdelacroix

PR #632, opened 2023-01-24
[MM-49700] Expose importer as a package

fmartingr:
#### Summary

Expose `commands/importer` as a package (moved to `pkg`) to allow other projects to use it as dependency.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49700


noxer:
There is no need to move the package, it's accessible from other packages by default. Since it's not under an `internal` package it's fine.

fmartingr:
@agnivade @noxer Thanks for your input! I may have been doing things the old way for too long I guess. Still, I can't seem to use the `importer` package when under the `commands` directory, on a fresh package:

```
go: finding module for package github.com/mattermost/mmctl/v6/commands/importer
testmmctl imports
        github.com/mattermost/mmctl/v6/commands/importer: module github.com/mattermost/mmctl/v6@latest found (v6.7.2), but does not contain package github.com/mattermost/mmctl/v6/commands/importer
```

I have checked the documentation and it doesn't explicitly specify that you can't use nested directories in packages nor that `commands` has any special behaviour, so I find this a bit weird, and one of the reasons I moved it to `pkg` in the first place.

fmartingr:
@agnivade @noxer Disregard this PR. After a few tests the problem was that I was using the `latest` package, but for golang that was v6.4.2  since the package name contains `/v6` instead of the v7. 

I've pinned it to the v7.7.1 commit and it works as expected now.

PR #666, opened 2023-06-05
Update README to point to monorepo

mgdelacroix:
#### Summary
Updates the README of the repository to point to the monorepo


sbishel (approved):

sbishel (approved):

isacikgoz (approved):
Is #666 PR number a coincidence 👁️

mgdelacroix:
@nab-77 historically we've been bad at keeping examples in the `README` up to date, so I think it would be better to directly point to the [mmctl web docs](https://docs.mattermost.com/manage/mmctl-command-line-tool.html). I can add a line in this PR referencing them and a `README` file in the mono-repo with an intro to the tool and a link to this docs so we have it in both places

PR #719, opened 2024-07-04
[chore] migration of plugin-store

phoinixgrr:
#### Summary

Plugin store is gradually migrated from:
- https://plugins-store.test.mattermost.com to
- https://plugins.releases.mattermost.com

We reflect that change here

Note: Currently both CDN's are working as expected, to facilitate the mgiration. Upon succesfull migration, https://plugins-store.test.mattermost.com will be decomissioned

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7958


mickmister (approved):

hanzei (changes requested):
This repo is no longer used. Instead, the test in https://github.com/mattermost/mattermost should get updated. That is the place where `mmctl` now lives.

phoinixgrr:
> This repo is no longer used. Instead, the test in https://github.com/mattermost/mattermost should get updated. That is the place where `mmctl` now lives.

Right. Any ideas why this is not Archived? Should we archive this? 
@hanzei

hanzei:
Yes, I think we are good to archive the repo. @mgdelacroix What do you think?

phoinixgrr:
> Yes, I think we are good to archive the repo. @mgdelacroix What do you think?

By the way, this is done with https://github.com/mattermost/mattermost/pull/27506

mgdelacroix:
@hanzei @phoinixgrr I agree, we should archive this repo

